PR #13, opened 2017-01-09
added task name to TaskResult

stuartaccent:
As per request, I have been asked to pr this, if you want me to do anything else with this please shout

Stu :)

PR #14, opened 2017-01-27
Add support for task name, args and kwargs

Photonios:
This allows you see which task failed and with what arguments it was invoked. In case this information is not available, it will simply not be stored.

This should probably be combined with the changes made in this PR: https://github.com/celery/django-celery-results/pull/13


django_celery_results/models.py
```diff
@@ -32,6 +32,18 @@ class TaskResult(models.Model):
...
+        null=True, max_length=255
+    )
+    task_args = models.TextField(
+        _('task arguments'),
+        null=True
+    )
+    task_kwargs = models.TextField(
+        _('task keyword arguments'),
+        null=True
+    )
```
jezdez: Following the [Django recommendation](https://docs.djangoproject.com/en/1.10/ref/models/fields/#null) for string based fields, this should not be using `null=True` but instead `default='', blank=True`. Then only set the values when storing the result if given, the default will take care of the rest.

setup.py
```diff
@@ -133,7 +133,7 @@ def run_tests(self):
 setuptools.setup(
     name=NAME,
     packages=setuptools.find_packages(exclude=['ez_setup', 't', 't.*']),
-    version=meta['version'],
+    version='1.0.4',
```
miigotu: I don't think this was meant to be in the PR?

thedrow:
This was fixed by #34.

PR #28, opened 2017-10-06
Make django-celery-results compatible with MySQL charset utf8mb4

Duologic:
Basically a reopen of https://github.com/celery/django-celery-results/pull/23


Duologic:
It is kinda backwards compatible. We simply found this fork and it looks like exactly what we wanted to do. On our own projects we did the change too (see django-triggers and django-db-locking). We estimated the impact of this change and figured it would only break some edge cases.

We hope you can apply this change so we don't have to start a fork.

I had a quick look at the errors and it looks unrelated to my change. I'll have a better look later.

auvipy:
send a clean PR

Issue #25, opened 2017-09-11
using Cache backends requires the app to be in INSTALLED_APPS


riccardodivirgilio:
I would like to use the CacheBackend but in order to do that I need to have this app registered as INSTALLED_APPS.
this means that migrate will be creating database table for DatabaseBackend which I'm not using.
this is happening because in django_celery_results.backends the module imports DatabaseBackend, and this is imported even if it's unused.

this triggers the import of TaskResult, which triggers an error from __new__ in django.

```
    from django_celery_results.backends.cache import CacheBackend as DjangoCacheBackend
  File "/usr/local/rotostampa/workflow-local/env/lib/python3.6/site-packages/django_celery_results/backends/__init__.py", line 4, in <module>
    from .database import DatabaseBackend
  File "/usr/local/rotostampa/workflow-local/env/lib/python3.6/site-packages/django_celery_results/backends/database.py", line 7, in <module>
    from ..models import TaskResult
  File "/usr/local/rotostampa/workflow-local/env/lib/python3.6/site-packages/django_celery_results/models.py", line 17, in <module>
    class TaskResult(models.Model):
  File "/usr/local/rotostampa/workflow-local/env/lib/python3.6/site-packages/django/db/models/base.py", line 113, in __new__
    "INSTALLED_APPS." % (module, name)
RuntimeError: Model class django_celery_results.models.TaskResult doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

can you remove the import of DatabaseBackend from there or remove the import of TaskResult?

thanks!

auvipy:
you could send a pr with your suggested changes

dylan-grafmyre:
I got this stacktrace when I had mistakenly not included "django_celery_results" in settings.py "INSTALLED_APPS", as documented (typo on my part)

PR #30, opened 2017-10-20
Check if django-cache used & Admin filters for celery results

wardal:
Check if django-cache used for results backend and don't import DatabaseBackend #25 .


wardal:
Sorry for 2 commits in 1 pull request.

PR #32, opened 2017-10-26
Check if django cache is used

wardal:
Check if django cache is used https://github.com/celery/django-celery-results/issues/25


thedrow:
Why is this needed?

auvipy:
dont thinks its useful so closing

PR #45, opened 2018-01-29
Use celery's b64decode and b64encode

D3X:
The original functions (from `base64`) return `bytes` objects on Python 3. Django 2 does not encode them to `str` before putting them in the database, resulting in strings like `"b'gAJ9cQBYCAAAAGNoaWxkcmVucQFdcQJzLg=='"` ending up in the database. Those in turn cannot be correctly `b64decode`-d and unpickled.

Using `celery`'s wrapper functions solves that problem.

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -1,8 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
-from base64 import b64encode, b64decode
-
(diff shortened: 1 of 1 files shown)


D3X:
@ask hi, sorry to bother you, but is there any chance you could review this? That bug is the last thing that prevents my company from upgrading to Django 2.

PR #54, opened 2018-06-27
Admin filters

jaylynch:
Some convenience additions to the list view to allow people to more easily skim through large result sets:

- Date hierarchy for browsing by date more easily over a longer period of time
- Filter by task name (eg. to see results of a daily task amongst a large list of those run every minute)
- Search by task details (eg. to find the results of a single item you know ran, or to search all "category.taskname failure" without clicking multiple filters)


jaylynch:

PR #55, opened 2018-06-27
Admin list

jaylynch:
Some minor tweaks to the list view that make life easier when there are results that tend to be clustered together in large quantities:

- Default ordering to date descending
- Have the first field / linked field in the list be the Task ID (when you click you are clicking to view that task, not anything about the task name, which is often repeated hundreds of times in a row)


jaylynch:

PR #56, opened 2018-06-27
Add ALLOW_EDITS Django setting

jaylynch:
Since this is primarily for viewing results, in our use case I don't really find it makes sense to allow people to edit the fields in the admin.

At best it will be confusing for regular users, I would like to encourage them to use this UI to provide more detail when reporting on one of their scheduled Beat tasks that has failed but don't want them to edit things.

I suppose some people might be using it as a testing tool or reprocessing results they edit the status of somehow though, so this adds a `DJANGO_CELERY_RESULTS` Django setting, with an `ALLOW_EDITS` key that can be set.

So you can add:

```python
DJANGO_CELERY_RESULTS: {
    'ALLOW_EDITS': False
}
```

to your Django config and it will make all fields read-only.

Default behaviour is unchanged.

### DIFF
--- a/django_celery_results/admin.py
+++ b/django_celery_results/admin.py
@@ -3,6 +3,13 @@ from __future__ import absolute_import, unicode_literals
 
 from django.contrib import admin
 
+from django.conf import settings
+try:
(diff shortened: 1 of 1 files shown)


jaylynch:

PR #63, opened 2018-08-31
Support specifying the database to use for the store_result method

chdsbd:
This change allows specifying the database to using when storing a result to the database.

I use this change to update a celery task while it is working on Django models in a transaction. With this change, I can use a separate connection to update the task without waiting for the transaction to commit.

I've changed the tests to depend on postgres to run. This is a requirement to be able to test this feature since sqlite seems to lock in a way that prevents multiple writers.

This change is related to https://github.com/celery/celery/pull/5017


chdsbd:
I fixed the CI problems and the related celery PR (https://github.com/celery/celery/pull/5017) has been merged. Is there anything I should do to get this merged?

chdsbd:
I'm not sure why this test doesn't work. I think it could be because we are using sqlite and not something like Postgres.

chdsbd:
Now I think it may be due to some locking/transaction issue that occurs when running the tests. I think it may have to due with how Django runs each test in a transaction and rolls it back at the end.

chdsbd:
@auvipy, I fixed the problems I was having with Django, but now Sqlite doesn't seem to be working well in this case. Sqlite seems to lock with the transaction and that prevents the second database connection from working.

I tried this with Postgres and all the tests pass. Should I put this test behind a flag and add Postgres to the testing setup and CI?

chdsbd:
Here is a diff to configure postgres: 
https://gist.github.com/chdsbd/c329ee831823c699a2246f0909b86cd2

auvipy:
thanks I will configure PostgreSQL

vogelier:
Wondering about the status here? Would be great to see this in action.

chdsbd:
@vogelier We need to move from sqlite to postgres for testing.

chdsbd:
@auvipy I've changed the tests to use postgres instead of sqlite. Let me know if I need to update anything else to get this merged.
I'm not sure why the apicheck test is failing. Could you point me in the right direction to fix that?

chdsbd:
It seems like pypy test is failing because of our psyopg2 version.

https://stackoverflow.com/questions/53338404/pypy3-import-psycopg2-error-undefined-symbol-pycoder-encoder

chdsbd:
I've added psycopg2cffi for pypy compatibility and merged in master. I'm not sure why the apicheck and python3.4 tests are failing, but I think it's unrelated to my change.

chdsbd:
@auvipy I've merged in master

auvipy:
thanks you. waiting for the travis signal to merge

chdsbd:
@auvipy Just fyi, this changes the tests to require postgres to run. Are you okay with that? If not, we probably could try to do some check to only run the tests when postgres is used.

t/proj/settings.py
```diff
@@ -77,12 +82,28 @@
...
+        'USER': 'postgres',
+        'HOST': 'localhost',
+        'PORT': '5432',
+        'OPTIONS': {
+            'connect_timeout': 1000,
+        }
+    },
+    'secondary': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': 'postgres',
```
liquidpele: Since your testing settings have two connections but to the same database, how do your tests verify that it wrote to the correct one?  Shouldn't you specify two different table names?

auvipy:
No problem at all with the current approach.

t/proj/settings.py
chdsbd: We verify that we wrote the database here: https://github.com/celery/django-celery-results/pull/63/files#diff-5083bf592c9b0d766e45e8c37e83db8eR98

The goal of this change is to allow updating the same database within a transaction.

liquidpele: Ah, I see now.  Neat.

PR #71, opened 2018-10-22
Fix MySQL max length issue.

milind-shakya-sp:
Current issues with mysql + `django-celery-results`

 If you want to run ``django-celery-results`` with MySQL, you might run into some issues.
 One such issue is when you try to run ``python manage.py migrate django_celery_results``, you might get the following error
```
     django.db.utils.OperationalError: (1071, 'Specified key was too long; max key length is 767 bytes')
```
 To get around this issue, with this pr, we can set the following in the settings file
```     
DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH=191
```
 (or any other value if any other db other than MySQL is causing similar issues.)
 max_length of **191** seems to work for MySQL.


milind-shakya-sp:
@auvipy added migration and tests. another 👀 ?

milind-shakya-sp:
@auvipy on second thoughts, an up date to the first migration will be required like it was initially done, since if not, the first migration will always fail on databases like mysql.

auvipy:
could you rebase on top of the master?

milind-shakya-sp:
@auvipy rebased from master

xuhuiding0612:
Hi guys,

Sorry for commenting under closed PR. One quick question would be is there a way to modify the max_length of existing 'django_celery_results_taskresult' table. (The table has already been created with default max_length 255)

I have added line 'DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH=191' in my settings. Just need app django_celery_results to trigger another migration for the max_length changes.

Issue #75, opened 2018-11-13
New app config changes the app_label for django_celery_results


LegoStormtroopr:
This will cause migrations to break if people had foreign keys to django_celery_results models.

LegoStormtroopr:
The specific error is:

```LookupError: No installed app with label 'django_celery_results'. Did you mean 'celery_results'?```

More details in #76

PR #76, opened 2018-11-13
Reverted app label to original default app_label #75

LegoStormtroopr:
From: https://docs.djangoproject.com/en/2.1/ref/applications/#django.apps.AppConfig.label

```
AppConfig.label¶
Short name for the application, e.g. 'admin'

This attribute allows relabeling an application when two applications have conflicting labels. It defaults to the last component of name. It should be a valid Python identifier.

It must be unique across a Django project.
```

Up until v1.0.2, `django-celery-results` didn't have an app label, so it defaulted to `django_celery_results`. The latest change breaks any instance of `django.apps.get_app_config` looking up via the original label, and will break all migrations that use the original label.

Any change we can get this reverted?

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.0.2
+current_version = 1.0.3
 commit = True
 tag = True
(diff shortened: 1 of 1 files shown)


auvipy:
@thedrow should we revert this

LegoStormtroopr:
Thanks very much, glad I could help!

Issue #61, opened 2018-07-27
New release?


xfxf:
Hi - now that Celery 4.2 is released, is it possible to do a new release?  I've been using a Celery 4.2 pre-release for a while and had to point to a later django-celery-results commit for it to work (4cc5f274b5d59d72f243e15d2d67c02a410b8b1d); I don't recall why (it wouldn't have been due to that specific commit; that was probably just the latest at the time I did it). 

Last release was quite some time ago.

jathanism:
Seconded. I'd really like the name/args/kwargs fields in `TaskResult`! It's been in the master branch for a while and I'm really trying to hold out on forking!

woegjiub:
Is there any ETA on this?
We've been specifying git-master in our Pipfile for a while now and it's been working perfectly, but it would be really nice to have a release on pypi to point to instead.

auvipy:
@thedrow

munavvar20:
Any ETA on the new release?

xblitz:
yes please, just a release!

marcomayer:
Yes, it's been a while since celery 4.2 was released and just using the latest master isn't a stable solution.

thedrow:
I'm really sorry this took so long.
1.0.2 is out :)

PR #36, opened 2017-10-27
Add support task_args, task_kwargs and task_name

checkpoint:
With this fix, detailed information on the task will be available. ( name, args, kwargs )
It based on with reference to the following PR. ( some improvements)
* https://github.com/celery/django-celery-results/pull/14/files

PR #78, opened 2018-11-14
Encode 'task_args' and 'task_kwargs' in 'store_result'.

arnau126:
Encode `task_args` and `task_kwargs` in `DatabaseBackend.store_result()` (as we do with `meta` and `result`) so we can retrieve them as python objects later on.

Note that this PR changes the serialization format of these fields.
For example, if `args=['a', 1, True, None]`
Currently it's stored as :
    `"['a', 1, True, None]"`
and after this PR it will be stored as:
    `'["a", 1, true, null]'` (json format)

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -21,11 +21,13 @@ class DatabaseBackend(BaseDictBackend):
             'children': self.current_task_children(request),
         })
 
-        task_name = getattr(request, 'task', None) if request else None
-        task_args = getattr(request,
-                            'argsrepr', getattr(request, 'args', None))
-        task_kwargs = getattr(request,
-                              'kwargsrepr', getattr(request, 'kwargs', None))
+        task_name = getattr(request, 'task', None)
+        _, _, task_args = self.encode_content(
+            getattr(request, 'argsrepr', getattr(request, 'args', None))
+        )
+        _, _, task_kwargs = self.encode_content(
+            getattr(request, 'kwargsrepr', getattr(request, 'kwargs', None))
+        )
         worker = getattr(request, 'hostname', None)
 
         self.TaskModel._default_manager.store_result(
@@ -46,8 +48,13 @@ class DatabaseBackend(BaseDictBackend):
         obj = self.TaskModel._default_manager.get_task(task_id)
         res = obj.as_dict()
         meta = self.decode_content(obj, res.pop('meta', None)) or {}
-        res.update(meta,
-                   result=self.decode_content(obj, res.get('result')))
+        result = self.decode_content(obj, res.get('result'))
+        task_args = self.decode_content(obj, res.get('task_args'))
+        task_kwargs = self.decode_content(obj, res.get('task_kwargs'))
+        res.update(
+            meta, result=result, task_args=task_args,
+            task_kwargs=task_kwargs,
+        )
         return self.meta_from_decoded(res)
 
     def encode_content(self, data):
--- a/t/unit/backends/test_database.py
+++ b/t/unit/backends/test_database.py
@@ -27,18 +27,36 @@ class test_DatabaseBackend:
             'django_celery_results.backends:DatabaseBackend')
         self.b = DatabaseBackend(app=self.app)
 
-    def test_backend__pickle_serialization(self):
+    def test_backend__pickle_serialization__dict_result(self):
         self.app.conf.result_serializer = 'pickle'
         self.app.conf.accept_content = {'pickle', 'json'}
         self.b = DatabaseBackend(app=self.app)
 
         tid2 = uuid()
         result = {'foo': 'baz', 'bar': SomeClass(12345)}
-        self.b.mark_as_done(tid2, result)
-        # is serialized properly.
-        rindb = self.b.get_result(tid2)
-        assert rindb.get('foo') == 'baz'
-        assert rindb.get('bar').data == 12345
+        request = mock.MagicMock()
+        request.task = 'my_task'
+        request.args = ['a', 1, SomeClass(67)]
+        request.kwargs = {'c': 6, 'd': 'e', 'f': SomeClass(89)}
+        request.hostname = 'celery@ip-0-0-0-0'
+        del request.argsrepr, request.kwargsrepr
+
+        self.b.mark_as_done(tid2, result, request=request)
+        mindb = self.b.get_task_meta(tid2)
+
+        assert mindb.get('result').get('foo') == 'baz'
+        assert mindb.get('result').get('bar').data == 12345
+        assert mindb.get('task_name') == 'my_task'
+
+        assert len(mindb.get('task_args')) == 3
+        assert mindb.get('task_args')[0] == 'a'
+        assert mindb.get('task_args')[1] == 1
+        assert mindb.get('task_args')[2].data == 67
+
+        assert len(mindb.get('task_kwargs')) == 3
+        assert mindb.get('task_kwargs')['c'] == 6
+        assert mindb.get('task_kwargs')['d'] == 'e'
+        assert mindb.get('task_kwargs')['f'].data == 89
 
         tid3 = uuid()
         try:
@@ -48,6 +66,66 @@ class test_DatabaseBackend:
 
         assert self.b.get_status(tid3) == states.FAILURE
         assert isinstance(self.b.get_result(tid3), KeyError)
+
+    def test_backend__pickle_serialization__str_result(self):
+        self.app.conf.result_serializer = 'pickle'
+        self.app.conf.accept_content = {'pickle', 'json'}
+        self.b = DatabaseBackend(app=self.app)
+
+        tid2 = uuid()
+        result = 'foo'
+        request = mock.MagicMock()
+        request.task = 'my_task'
+        request.args = ['a', 1, SomeClass(67)]
+        request.kwargs = {'c': 6, 'd': 'e', 'f': SomeClass(89)}
+        request.hostname = 'celery@ip-0-0-0-0'
+        del request.argsrepr, request.kwargsrepr
+
+        self.b.mark_as_done(tid2, result, request=request)
+        mindb = self.b.get_task_meta(tid2)
+
+        assert mindb.get('result') == 'foo'
+        assert mindb.get('task_name') == 'my_task'
+
+        assert len(mindb.get('task_args')) == 3
+        assert mindb.get('task_args')[0] == 'a'
+        assert mindb.get('task_args')[1] == 1
+        assert mindb.get('task_args')[2].data == 67
+
+        assert len(mindb.get('task_kwargs')) == 3
+        assert mindb.get('task_kwargs')['c'] == 6
+        assert mindb.get('task_kwargs')['d'] == 'e'
+        assert mindb.get('task_kwargs')['f'].data == 89
+
+    def test_backend__pickle_serialization__bytes_result(self):
+        self.app.conf.result_serializer = 'pickle'
+        self.app.conf.accept_content = {'pickle', 'json'}
+        self.b = DatabaseBackend(app=self.app)
+
+        tid2 = uuid()
+        result = b'foo'
+        request = mock.MagicMock()
+        request.task = 'my_task'
+        request.args = ['a', 1, SomeClass(67)]
+        request.kwargs = {'c': 6, 'd': 'e', 'f': SomeClass(89)}
+        request.hostname = 'celery@ip-0-0-0-0'
+        del request.argsrepr, request.kwargsrepr
+
+        self.b.mark_as_done(tid2, result, request=request)
+        mindb = self.b.get_task_meta(tid2)
+
+        assert mindb.get('result') == b'foo'
+        assert mindb.get('task_name') == 'my_task'
+
+        assert len(mindb.get('task_args')) == 3
+        assert mindb.get('task_args')[0] == 'a'
+        assert mindb.get('task_args')[1] == 1
+        assert mindb.get('task_args')[2].data == 67
+
+        assert len(mindb.get('task_kwargs')) == 3
+        assert mindb.get('task_kwargs')['c'] == 6
+        assert mindb.get('task_kwargs')['d'] == 'e'
+        assert mindb.get('task_kwargs')['f'].data == 89
 
     def xxx_backend(self):
         tid = uuid()


auvipy:
Is this a backward incompatible change?

arnau126:
I think that this change won't break any main functionality, as these fields are just informative or for debugging.
As I said the difference is the way these fields are stored in the DB:
    `"['a', 1, True, None]"` vs. `'["a", 1, true, null]''`

For me it's not a problem that the storing format change.

Maybe it's a problem for someone who has a script or something that parse and use these fields. However, these fields are quiet new (the deploy was 2 days ago).

kiyoya:
Could we make it an option?

Yes, the new release was landed in a few days ago. However the feature was merged several months ago (#36) and not a few people have been using `HEAD` instead of 1.0.1 (#61).

arnau126:
I don't like the idea to make it an option. I think that not encoding these fields is a bug, because serializing a `list` or a `dict` using `str()` force you to use `eval()` to retrieve them as python objects, which is not safe.

kiyoya:
It's a bit off-topic but there is `ast.literal_eval` for parsing strings like `task_args`.

Yes, I agree that`task_args` and `task_kwargs` should have been implemented like this in the first place. Unfortunately they weren't, however, and I'm not sure if it is worth making an incompatible change while there is a safe way to parse them.

Another possibility is to introduce new fields in parallel.

django_celery_results/backends/database.py
```diff
@@ -41,8 +43,13 @@ def _get_task_meta_for(self, task_id):
         obj = self.TaskModel._default_manager.get_task(task_id)
         res = obj.as_dict()
         meta = self.decode_content(obj, res.pop('meta', None)) or {}
-        res.update(meta,
-                   result=self.decode_content(obj, res.get('result')))
+        result = self.decode_content(obj, res.get('result'))
+        task_args = self.decode_content(obj, res.get('task_args'))
+        task_kwargs = self.decode_content(obj, res.get('task_kwargs'))
```
liquidpele: The args and kwargs encoding shouldn't match the result encoding, which is what this logic essentially does since you pass in obj and decode_content checks obj.content_encoding.  You ignore the returned content_encoding for those items above, so it's basically lost information.   Result encoding can be binary for instance, but args/kwargs may not be, and vice versa.

t/unit/backends/test_database.py
```diff
@@ -33,11 +35,19 @@ def test_backend__pickle_serialization(self):
...
         result = {'foo': 'baz', 'bar': SomeClass(12345)}
-        self.b.mark_as_done(tid2, result)
-        # is serialized properly.
-        rindb = self.b.get_result(tid2)
-        assert rindb.get('foo') == 'baz'
-        assert rindb.get('bar').data == 12345
+        request = MagicMock()
+        request.task = 'my_task'
+        request.args = ['a', 1]
+        request.kwargs = {'c': 3, 'd': 'e'}
```
liquidpele: Are these the only types of inputs that can be received?  What about other things?  

You really need to look at all the different things that the kombu dumps function (which is what actually handles...  

https://github.com/celery/kombu/blob/master/kombu/serialization.py#L161

django_celery_results/backends/database.py
arnau126: I use `self.decode_content(...)` to deserialize because I used `self.encode_content(...)` to serialize. I thought that it was consistent. I don't know any other way to do it.

t/unit/backends/test_database.py
arnau126: I think that `args` is always a list and `kwargs` always a dict.
So, you mean something like:
```
request.args = ['a', 1, ['2', 3, {'b': 'x'}], {'c': 'd', 'e': 3, 'f': [5, 6, 7]}]
```
?

liquidpele: Yes, but an array/dict containing what?

django_celery_results/backends/database.py
liquidpele: Look at how the decode works.

arnau126: Look at how encode works.

t/unit/backends/test_database.py
arnau126: The `content_encoding` that returns the `encode_content` function only depends on `self.serializer`. Which means that it will be the same for `result`, `meta`, `task_args`, `task_kwargs` and for all others objects serialized using the `encode_content` of the same DatabaseBackend instance.

Moreover, we're also ignoring the `content_encoding` of `meta` and there is error when we deserialize it using the same `decode_content` function.

django_celery_results/backends/database.py
arnau126: The `content_encoding` that returns the `encode_content` function only depends on `self.serializer`. Which means that it will be the same for `result`, `meta`, `task_args`, `task_kwargs` and for all others objects serialized using the `encode_content` of the same DatabaseBackend instance.

Moreover, we're also ignoring the `content_encoding` of `meta` and there is error when we deserialize it using the same `decode_content` function.

t/unit/backends/test_database.py
arnau126: I don't get you.
Is it okay or not this example: `request.args = ['a', 1, ['2', 3, {'b': 'x'}], {'c': 'd', 'e': 3, 'f': [5, 6, 7]}]` ?
Let me know which other data types I could include.

arnau126:
Rebased

auvipy:
> Rebased

let's reach a consensus for the changes before the merge.

arnau126:
I'm pretty sure that the code works.
The important issue here is the backward incompatibility, which @kiyoya and I already discussed.
I'm willing to hear other points of view.

auvipy:
Keeping BC is very important.

arnau126:
So the options are:
- Make it an option (maybe a setting).
- Introduce new fields in parallel.
- Wait until a major release where backward incompatibility are allowed.
- Don't do it at all (reject the PR).

Let me know if you come up with other ideas.

t/unit/backends/test_database.py
liquidpele: What I mean, is what if I pass an object as an arg?  What about somethign that's not json serializable like django's lazy gettext objects?  What about a file handle?  A string that's 1 million bytes?  etc.  Your tests handle the easy test cases, but those are the kinds of tests that aren't particularly helpful... try to break your own code ;)

django_celery_results/backends/database.py
liquidpele: obj.content_encoding is set according to result, but encoding doesn't take that into account, it only looks at the data for the field in question.  What you've done is made a case where the result may be binary but it encoded your args as json... but then trying to decode the args would try to decode the binary which it wasn't encoded as.  

Basically content_encoding was only ever meant for result.  If you want to encode the args/kwargs you need to make sure it always handles them the same way, otherwise there will be a lot of inconsistency.

arnau126: > obj.content_encoding is set according to result

False, it is not set according to `result`. It is set according to `self.serializer`, which is always the same.

Kombu `dumps()` function would only take `result` into account if `self.serializer` were `None` or `"raw"`. But this is not possible because celery's allowed serializers are only: json, pickle, yaml and msgpack (or custom serializer, but not None nor "raw").

liquidpele: Hmm, maybe I'm reading this wrong.  What is the obj args in decode_content?  I was assuming that was the django model instance.

arnau126: I added more test where:
- result is a dict
- result is text
- result is bytes

to ensure that it works fine with all of them.

t/unit/backends/test_database.py
arnau126: I included the custom testing class `SomeClass` in args and kwargs to make sure that it is able to serialize non standard objects.

liquidpele: awesome thanks!

django_celery_results/backends/database.py
arnau126: > Hmm, maybe I'm reading this wrong. What is the obj args in decode_content? I was assuming that was the django model instance.

The `obj` argument of the `decode_content()` function is a `TaskResult` instance.

liquidpele: ah, okay, then this makes much more sense, sorry about the confusion.  If all your extra tests are passing then I'm good besides @auvipy 's concern about backward compatibility.

auvipy:
> So the options are:
> 
> * Make it an option (maybe a setting).
> * Introduce new fields in parallel.
> * Wait until a major release where backward incompatibilities are allowed.
> * Don't do it at all (reject the PR).
> 
> Let me know if you come up with other ideas.

I would wait for celery 5 and this package 2.0 version

KevinPoole:
> It's a bit off-topic but there is `ast.literal_eval` for parsing strings like `task_args`.
> 
> Yes, I agree that`task_args` and `task_kwargs` should have been implemented like this in the first place. Unfortunately they weren't, however, and I'm not sure if it is worth making an incompatible change while there is a safe way to parse them.
> 
> Another possibility is to introduce new fields in parallel.

Looks like this particular PR has already been settled as "not yet" but I just wanted to comment on the above: `ast.literal_eval` is insufficient for cases where, for example, `pickle` has been used for serialization and there exists some `CustomClass` objects in the kwargs. In this case `ast.literal_eval` chokes `invalid syntax` when you get to the `repr` string for the objects.

ZviBaratz:
Is there anything I can do to help move this forward?

arnau126:
Hope that now that Celery 5 is released, it's a good moment to update this package to version 2.0, and merge this PR.

auvipy:
is it possible to not to break existing systems?

arnau126:
This PR will only change how new TaskResult instances will store `task_args` and `task_kwargs`. Already existing instances will keep the current format.
So the problem is that we will have old instances with these fields serialized using `str()` and new instances with these fields serialized using `json.dumps()`

I already suggested a few options:
> - Make it an option (maybe a setting).
> - Introduce new fields in parallel.
> - **Wait until a major release where backward incompatibilities are allowed.**
> - Don't do it at all (reject the PR).

However, I think that store a `list` or a `dict` using `str()` is not a good practice, so it shouldn't be an option at all.
That's why I would wait until a major release.
And I think that now is a good moment to do it: celery 5 is released, py35 support has been dropped...

johnthealy3:
I just encountered this after solving a similar problem trying to use the JSONField (I was using the postgres contrib one but now merged in Django 3.1: https://docs.djangoproject.com/en/3.1/ref/contrib/postgres/fields/#jsonfield) for task_args and task_kwargs. Perhaps now is the time, with an option or with Django version checking?

Happy to provide some code if this approach has potential.

ZviBaratz:
@arnau126 I tried to check out this PR and I'm still getting `task_args` and `task_kwargs` encoded as strings. Any idea what I'm missing?

arnau126:
`task_args` and `task_kwargs` still are `TextField` fields, so when you retrieve a TaskResult instance from DB they are strings. To cast them to `list` or `dict` you have to do `json.loads`

ZviBaratz:
@arnau126 that makes sense, sorry :sweat_smile:

auvipy (approved):

auvipy:
I am accepting this PR, but new fields / better solutions using django new fields etc is wellcome

robvdl:
So 2.0.0 didn't fix the CVE yet? issue #154 when I saw a new major release I had my hopes up the CVE was fixed but I don't think it has yet. Has version 2.0.0 been flagged yet or is it a matter of time that version gets flagged too?

arnau126:
@robvdl , this PR wasn't meant to fix #154 at all, but to change the way the input arguments were stored (use `json.dumps` instead of `str`).

igorvoltaic:
> Encode `task_args` and `task_kwargs` in `DatabaseBackend.store_result()` (as we do with `meta` and `result`) so we can retrieve them as python objects later on.
> 
> Note that this PR changes the serialization format of these fields.
> For example, if `args=['a', 1, True, None]`
> Currently it's stored as :
> `"['a', 1, True, None]"`
> and after this PR it will be stored as:
> `'["a", 1, true, null]'` (json format)

Seems like not really working. This is what i need to do to get back the task arguments.
```
>>> from django_celery_results.models import TaskResult
>>> import ast
>>> import json
>>> TaskResult.objects.last()
<TaskResult: <Task: c4e59e69-d324-44fd-95ad-1a8c210fa2bd (SUCCESS)>>
>>> t = TaskResult.objects.last()
>>> t.task_args
'"({\'dataset_id\': 2, \'height\': 2600, \'width\': 2600, \'plot_type\': \'bar\', \'params\': {\'x\': \'month\', \'y\': \'passengers\', \'hue\': None}, \'columns\': [\'month\', \'passengers\'], \'id\': 1},)"'
>>> json.loads(t.task_args)
"({'dataset_id': 2, 'height': 2600, 'width': 2600, 'plot_type': 'bar', 'params': {'x': 'month', 'y': 'passengers', 'hue': None}, 'columns': ['month', 'passengers'], 'id': 1},)"
>>> ast.literal_eval(json.loads(t.task_args))
({'dataset_id': 2, 'height': 2600, 'width': 2600, 'plot_type': 'bar', 'params': {'x': 'month', 'y': 'passengers', 'hue': None}, 'columns': ['month', 'passengers'], 'id': 1},)

>>> ast.literal_eval(json.loads(t.task_args))[0]
{'dataset_id': 2, 'height': 2600, 'width': 2600, 'plot_type': 'bar', 'params': {'x': 'month', 'y': 'passengers', 'hue': None}, 'columns': ['month', 'passengers'], 'id': 1}
```

Do I correctly understand this commit was aimed to allow simply do `json.loads(t.task_args)` ?

auvipy:
this is a more recent PR https://github.com/celery/django-celery-results/pull/178 and @arnau126 opened a new PR in celery as well https://github.com/celery/celery/pull/6595

arnau126:
> Do I correctly understand this commit was aimed to allow simply do json.loads(t.task_args) ?

Yes, you understand correctly.


> Seems like not really working

You're right, unfortunately not working, because of the new Celery [task_protocol=2](https://docs.celeryproject.org/en/stable/userguide/configuration.html#std-setting-task_protocol) and the [hide-sensitive-information feature](https://docs.celeryproject.org/en/stable/userguide/tasks.html#hiding-sensitive-information-in-arguments)

I have a work-in-progress PR (celery/celery#6595) to fix it (to allow simply do `json.loads(t.task_args)`).

Issue #79, opened 2018-12-07
delete_expired does not send deletion related signal


pptime:
function **django_celery_results/managers.py: delete_expired** does not send django standard deletion related signal, i.e. **pre_delete**, **post_delete**

This may lead to potential errors if the TaskResult model is referenced as a ManyToManyField.

auvipy:
fixes are welcome

pptime:
I will try

PR #80, opened 2018-12-10
use queryset's deletion method instead

pptime:
To resolve issue #79 using the delete() method of a queryset.
All unit tests are passed. Please review if there is any side effect.


pptime:
when do you plan to release?

Issue #83, opened 2018-12-17
Unknown System Variable 'tx_isolation' when using MySQL 8


Planswalker:
Hello,

I recently have created a set of celery tasks using django_celery_results on a Mysql 8 backend. Most of the time this works well, but recently I've seen the following error while waiting for celery tasks to complete:

`_mysql_exceptions.OperationalError: (1193, "Unknown system variable 'tx_isolation'")`

When tracing the error I was able to locate the following in managers.py:

```   
 def warn_if_repeatable_read(self):
        if 'mysql' in self.current_engine().lower():
            cursor = self.connection_for_read().cursor()
            if cursor.execute('SELECT @@tx_isolation'):
                isolation = cursor.fetchone()[0]
                if isolation == 'REPEATABLE-READ':
                    warnings.warn(TxIsolationWarning(W_ISOLATION_REP.strip()))
```

Which I am pretty sure is the cause of this issue - `tx_isolation` was deprecated in MySQL 5.7 ([see here](https://dev.mysql.com/doc/refman/5.7/en/added-deprecated-removed.html#optvars-deprecated)) and MySQL 8 changed the name of the system variable from `tx_isolation` to `transaction_isolation` ([see here](https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html#mysql-nutshell-removals))

Thanks!

liquidpele:
Fix was merged, closing this.   Please re-open if the issue persists.

ndido98:
The PR responsible for fixing this issue has introduced a regression in MariaDB (I use 10.3.15, but I suppose this is valid for versions newer than 10.1), where the `transaction_isolation` variable doesn't exist and it's still named the old MySQL way, so `tx_isolation`.

Django itself had the same issue and the same regression, which [was fixed in this commit.](https://github.com/django/django/commit/e3c852cbd609484b272f563f3c21066fb12ef7f8).

liquidpele:
@Planswalker or @ndido98 could one of you make a PR with a similar fix that will handle both?

liquidpele:
Closing this since a PR was merged to fix the fix.

PR #84, opened 2018-12-18
Fix MySQL8 system variable issue

Planswalker:
This should fix the issue mentioned in #83 by using the correct system variable depending on the version of MySQL being used.

### DIFF
--- a/django_celery_results/managers.py
+++ b/django_celery_results/managers.py
@@ -133,7 +133,7 @@ class TaskResultManager(models.Manager):
     def warn_if_repeatable_read(self):
         if 'mysql' in self.current_engine().lower():
             cursor = self.connection_for_read().cursor()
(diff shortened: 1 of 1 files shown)


Planswalker:
I suppose it's possible, but in order to verify that the code works you'd need to test in both a version of mysql above 5.7.20 and below - I looked at the current testing framework and it doesn't seem like any mysql instance is being set up so I'm not sure if the current testing environment would be able to support it.

liquidpele:
Based on the docs this looks correct.   https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_tx_isolation 

Will go ahead and merge.   Just an as FYI to this thread, mysql 5.7.20 was released in Oct 2017. 
  https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-20.html

PR #94, opened 2019-05-05
Try to expand the matrix more, reconfiguring to let travis handle the…

liquidpele:
… matrix generation more for us.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
     - { python: pypy, env: DJANGO=2.1 }
     - { python: pypy, env: DJANGO=2.2 }
 
-install: travis_retry pip install -U tox | cat
+install: travis_retry pip install -U tox-travis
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,39 +1,44 @@
 language: python
 sudo: True
 cache: pip
-python:
-    - '3.6'
 os:
     - linux
 dist: xenial
+python:
+  - 2.7
```
auvipy: this is django 1.11 only so includ in django 1.11 was enough

liquidpele: what do you mean?

auvipy: Ask for review before merging from now on.

liquidpele: Sure, I can do that, but what do you mean by "this is django 1.11 only so includ in django 1.11 was enough" ?

PR #98, opened 2019-05-23
Fix 0004 migration

arnau126:
Fix 0004 migration, use `DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH` setting in the `task_id.max_length` instead of hardcoding `191` value.

### DIFF
--- a/django_celery_results/migrations/0004_auto_20190516_0412.py
+++ b/django_celery_results/migrations/0004_auto_20190516_0412.py
@@ -6,6 +6,7 @@
 
 from __future__ import absolute_import, unicode_literals
 
+from django.conf import settings
 from django.db import migrations, models
 
 
@@ -59,7 +60,17 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='taskresult',
             name='task_id',
-            field=models.CharField(db_index=True, help_text='Celery ID for the Task that was run', max_length=191, unique=True, verbose_name='Task ID'),
+            field=models.CharField(
+                db_index=True,
+                help_text='Celery ID for the Task that was run',
+                max_length=getattr(
+                    settings,
+                    'DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH',
+                    255
+                ),
+                unique=True,
+                verbose_name='Task ID'
+            ),
         ),
         migrations.AlterField(
             model_name='taskresult',


liquidpele:
I really dont like using a setting in a migration like this.  Sure it works, but I think it will only cause issues down the road.  

Same with the other PR like this, my question is can task_id even be anything other than a uuid?

arnau126:
> any test case? or related code that shows this change fix the issue?

I don't know how to test it. I just copied what @milind-shakya-sp did here https://github.com/celery/django-celery-results/commit/01a23a55ef33f19b6ea76d368ede9873db06223a.
When I updated django-celery-result to 1.1.0 I got this message: `Your models have changes that are not yet reflected in a migration, and so won't be applied.` With the changes of this PR, I don't get this message any more, and everything works as always.

> I really don't like using a setting in a migration like this.

Me neither. But currently in `master` is done this way, so I just copied the same pattern.


What is sure is that the new version (1.1.0) doesn't work for people who has a `max_length != 191`

liquidpele:
I'm fine with merging this as this is already the way it's done in this project, unfortunately.  If the task_id can only be a uuid though, we should just switch the thing to be a UUIDField  and the length becomes a moot point.

milind-shakya-sp:
> I really don't like using a setting in a migration like this.

Not really sure why some people "don't like" using a configurable setting in migration when the situation calls for it. There are many projects that follow this pattern, one such case is https://github.com/python-social-auth/social-app-django/blob/master/social_django/migrations/0001_initial.py#L13

auvipy:
> I really dont like using a setting in a migration like this. Sure it works, but I think it will only cause issues down the road. Edit: oh, it looks like this is already in use here... but this is a good example of the issues it causes.
> 
> Same with the other PR like this, my question is can task_id even be anything other than a uuid?

I don't see any problem

auvipy:
> I really dont like using a setting in a migration like this. Sure it works, but I think it will only cause issues down the road. Edit: oh, it looks like this is already in use here... but this is a good example of the issues it causes.
> 
> Same with the other PR like this, my question is can task_id even be anything other than a uuid?

I don't see any problem 

> I'm fine with merging this as this is already the way it's done in this project, unfortunately. If the task_id can only be a uuid though, we should just switch the thing to be a UUIDField and the length becomes a moot point.

I was suspicious after your pr as  model changes didn't trigger migration changes!!

auvipy:
> I really dont like using a setting in a migration like this. Sure it works, but I think it will only cause issues down the road. Edit: oh, it looks like this is already in use here... but this is a good example of the issues it causes.
> 
> Same with the other PR like this, my question is can task_id even be anything other than a uuid?

I don't see any problem 

> I'm fine with merging this as this is already the way it's done in this project, unfortunately. If the task_id can only be a uuid though, we should just switch the thing to be a UUIDField and the length becomes a moot point.

I was suspicious after your pr as  model changes didn't trigger migration changes!!

auvipy:
will push a minor release

auvipy:
will push a minor release

NyanKiyoshi:
@auvipy any news on a minor release for that?

auvipy:
pushing one now!!

PR #99, opened 2019-05-29
Fix transaction_isolation regression in MariaDB

ndido98:
Fixes again #83, since #84 introduced a regression in MariaDB, which as of 10.3 still uses the old `tx_isolation` name.


auvipy:
make sure this wont create another regression!

auvipy:
ping me if i forget to publish a minor release

ndido98:
Ping!

auvipy:
thanks for the ping

PR #108, opened 2019-07-25
Added worker as new task field

agutierrezrodriguez:
With this PR we can track the worker that execute the task too.


liquidpele:
Cool, this looks very useful.

Issue #87, opened 2019-02-06
Request to add "date created" to TaskResult model


steresi:
The TaskResult Model (e.g. stored in django_celery_results_taskresult) has a "date_done" field, but I recommend it should also have a "date_created". Since the record will be modified when the task is completed, the "date_done" may be different from the creation date.

A creation date would be very useful for checking when a task was submitted vs. when it was actually completed.

auvipy:
you can come with a pr with new migrations

ZviBaratz:
@auvipy can this be closed? Seems to be merged and working :+1:

bigott-a:
@auvipy I might have an issue with the field `date_created`

I've created a sample to reiterate it:

task.py
```python
@shared_task(bind=True, name="default_task")
def default_task(self, files=None):
    start = str(timezone.now())
    sleep(10)
    end = str(timezone.now())
    return f"{start=}\n{end=}"
```

The task is well executed and returned:
```shell
Task default_task[0dd3e0e8-b967-4b47-a507-1924f9383534] succeeded in 10.048574910964817
s: 
start = '2022-02-16 07:28:49.910870'
end = '2022-02-16 07:28:59.931319'
'
```

But if I checked the content the table `task_result`, date_created is nearly equal to the date_done:
date_done = 2022-02-16 14:28:59.95575+01
date_created = 2022-02-16 14:28:59.9554+01

I was expecting to get the start time of task in date_created.
Is it normal ?

auvipy:
probably it completes very fast?

PR #111, opened 2019-08-21
Added date_created field to TaskResult model.

hramezani:
https://github.com/celery/django-celery-results/issues/87


django_celery_results/models.py
```diff
@@ -63,6 +63,10 @@ class TaskResult(models.Model):
         verbose_name=_('Result Data'),
         help_text=_('The data returned by the task.  '
                     'Use content_encoding and content_type fields to read.'))
+    date_created = models.DateTimeField(
+        auto_now_add=True, db_index=True,
+        verbose_name=_('Created DateTime'),
+        help_text=_('Datetime field when the task was created in UTC'))
```
liquidpele: Is this when the task is created, or when the task result is created?

django_celery_results/models.py
```diff
@@ -63,6 +63,10 @@ class TaskResult(models.Model):
         verbose_name=_('Result Data'),
         help_text=_('The data returned by the task.  '
                     'Use content_encoding and content_type fields to read.'))
+    date_created = models.DateTimeField(
```
liquidpele: I feel like it would be a lot safer to set null=True, otherwise it's too easy to throw an error during migration if something doesn't get a value set.

liquidpele: Oh, I guess date_done doesn't allow nulls either.  Okay, this is probably fine then.

django_celery_results/models.py
hramezani: Yes, it is. I changed the help text.

hramezani:
@liquidpele Thanks for the comment. It is done now!

django_celery_results/models.py
liquidpele: Cool, thanks for the contribution!

PR #118, opened 2019-10-22
Delete results immediately, not in two steps

### LINKED ISSUE #117: Cleanup task times out
The `celery.backend_cleanup` task is super slow reaching our configured soft time limit of 15 minutes. The slow query is the delete and update queries called by `TaskResultManager.delete_expired()`:

https://github.com/celery/django-celery-results/blob/64315751dbe7eac8b3276acb8ed56db2add71b74/django_celery_results/managers.py#L173-L177

We can increase the limit for the cleanup task (by updating the `soft_time_limit` property of the task), but that will only be a bandaid rather than a long term solution.

I have reproduced a simplified case of this issue by creating 2mil results locally (only setting uuid() and using default of the model), setting 1mil to an older date and running the query:

```python
>>> from datetime import datetime, timedelta
>>> from django.db import connection
>>> from celery import states, uuid
>>> from django_celery_results.models import  *
>>> [TaskResult(task_id=uuid()) for _ in range(1_000_000)]
>>> [TaskResult(task_id=uuid()) for _ in range(1_000_000)]
>>> othermil = TaskResult.objects.filter(id__gt=1_000_000)

>>> old_date = datetime.now() - timedelta(days=5)
>>> othermil.update(date_done=old_date)

>>> print("\n".join(f"{q['sql'][:300]}: {q['time']}" for q in connection.queries))

INSERT INTO "django_celery_results_taskresult" ("task_id", "task_name", "task_args", "task_kwargs", "status", "worker", "content_type", "content_encoding", "result", "date_created", "date_done", "traceback", "hidden", "meta") VALUES ('33efa9af-08b3-4dc6-8808-4597f1da2b17', NULL, NULL, NULL, 'PEND...: 203.467
INSERT INTO "django_celery_results_taskresult" ("task_id", "task_name", "task_args", "task_kwargs", "status", "worker", "content_type", "content_encoding", "result", "date_created", "date_done", "traceback", "hidden", "meta") VALUES ('0f9d897a-bebb-42b3-ba24-f80061c97c95', NULL, NULL, NULL, 'PEND...: 214.064
UPDATE "django_celery_results_taskresult" SET "date_done" = '2019-10-16T13:19:12.314815+00:00'::timestamptz WHERE "django_celery_results_taskresult"."id" >= 1000002: 137.222

# and after running: TaskResult.objects.delete_expired(86_400)

UPDATE "django_celery_results_taskresult" SET "hidden" = true WHERE "django_celery_results_taskresult"."date_done" < '2019-10-20T18:41:52.125578+00:00'::timestamptz: 149.066
DELETE FROM "django_celery_results_taskresult" WHERE "django_celery_results_taskresult"."hidden" = true: 3.387
```

The UPDATE query takes over 2 minutes, the DELETE is pretty fast.

In our production case we're talking slightly bigger dataset than this, with significantly more data in other fields (e.g metadata of `chord_unlocks` to collect multiple tasks that were chained into a group). I just timed the "update portion" of the `delete_expired` task manually, and it took over an hour and a half:
```python
In [20]: %time expired_update = TaskResult.objects.get_all_expired(expires).update(hidden=True)
CPU times: user 12 ms, sys: 0 ns, total: 12 ms
Wall time: 1h 33min 46s
In [22]: TaskResult.objects.filter(hidden=True).delete()
Out[22]: (1977820, {'django_celery_results.TaskResult': 1977820})
```

Is it there anything we can do to help debug the issue, what should we consider to improve the database operation? Can we consider removing the index on mtime, or use a partial index maybe?
count

asfaltboy:
Inspiration by the benchmark in the related ticket, this change removes the slowest running operation in the cleanup task.

The same, direct deletion operation, is done in [the built-in database backend of celery implementation of the cleanup function][1].

To test, creates some records, and run clean up task:

```python
from datetime import datetime, timedelta
from django.db import connection
from celery import states, uuid
from django_celery_results.models import  *

def create_old_results(records=200000, age=5):
    """ takes 40s to create 200k old records on my machine """
    old_date = datetime.now() - timedelta(days=age)
    bunch = [TaskResult(task_id=uuid()) for _ in range(records)]
    created = TaskResult.objects.bulk_create(bunch)
    TaskResult.objects.filter(id__gte=created[0].id).update(date_done=old_date)
```

Before the PR
> Task celery.backend_cleanup[4cd0ad59-f241-4d3b-b848-45dfc9564725] succeeded in 8.506982273000176s: None

After the PR
> Task celery.backend_cleanup[069fe103-f894-4b96-ad1a-720a6bbbec6f] succeeded in 0.3088349770041532s: None

this somewhat fixes #117

I hesitate to add the test case above to the test suite as creating a large amount of records take a long while... and without it being large enough, the test may be too flaky and raise false positives (depending on the test env).

[1]: https://github.com/celery/celery/blob/c780e3a954579ee5b7243b9cb7444e44a6398d5b/celery/backends/database/__init__.py#L207-L217


asfaltboy:
> can you also add tests for this?

There's already a test that deletes expired results, and it seems it's passing. As I mentioned in the description, adding another test will add a minute to the test suite... in effect doubling the time it takes the tests to run (imagine waiting 30min for tox/travis to finish testing all envs).

From my experience, an integration test flag is useful for these cases, I can add an integration test marker + pytest argument that will skip the integration tests by default, and then run them only on a single env in tox & travis. What do you say @auvipy ?

On a different note, I think I will also remove the `TaskResult.hidden` field.

auvipy:
please move forward.

tox.ini
```diff
@@ -64,4 +65,10 @@ commands =
 usedevelop = true
 commands = pip install -U celery==4.3
            pip install -U kombu==4.5
-           py.test --cov=django_celery_results --cov-report=xml --no-cov-on-fail
+           pytest --cov=django_celery_results --cov-report=xml --no-cov-on-fail
+
+[testenv:integration]
+commands =
+    pip install -U celery==4.3
```
auvipy: try celery==4.4.0rc3

auvipy: kombu 4.6.x

asfaltboy: We can do that, but then, shouldn't we use the latest version for all test testenv builds, not just integration?

The easiest way to enable pre-release to be installed for all unpinned dependencies is to use the `pip_pre` tox flag: https://tox.readthedocs.io/en/latest/config.html#conf-pip_pre

PR #122, opened 2020-01-08
add task status api from django-celery

### LINKED ISSUE #103: any plans or interest in replicating the task status json response from django-celery?
I'm wondering if the omission of the django-celery feature for returning json data about task status in this project is intentional or not. Seems like this is the logical place for that functionality. If it's welcome I would be glad to add it but if it's omitted intentionally I will roll my own replacement and live with that.

nthall:
This adds the task status API from `django-celery` into the new module (resolves #103 ). Much of the code is copy-paste from the old library, but some changes were necessary for compatibility with newer versions Celery et al. For that reason it was easier to write new tests than adapt the old ones, but of course if that's not OK I can do it the other way.

In my local environment, all tests pass except for `apicheck`, which fails for me because my environment still defaults to python2.7, but `sphinx-celery` requires a version of `sphinx` that only supports python3.5 or greater. 

If I've missed anything or need to clean anything up I'll be happy to!

### DIFF
--- a/django_celery_results/urls.py
+++ b/django_celery_results/urls.py
@@ -1,0 +1,27 @@
+"""URLs defined for celery.
+
+* ``/$task_id/done/``
+    URL to :func:`~celery.views.is_successful`.
+* ``/$task_id/status/``
+    URL  to :func:`~celery.views.task_status`.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.conf.urls import url
+
+from . import views
+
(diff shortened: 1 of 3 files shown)


nthall:
Hm, on Travis's page the build has finished, with an error in pydocstyle pointing at something unrelated to this PR. I dunno what if anything to do about that.

auvipy:
django_celery_results/managers.py:96 in public method `store_result`:

        D417: Missing argument descriptions in the docstring (argument(s) meta, traceback are missing descriptions in 'store_result' docstring)

PR #124, opened 2020-01-14
Admin site - add worker to list_filter

maxmalysh:
Screenshot:

PR #125, opened 2020-01-14
Make TaskResults read-only by default

maxmalysh:
Django-Celery-Results is meant for monitoring. There is little to no reason to allow editing task results. So it would be better to allow edits only if a user explicitly specifies this setting. This is how it worked in django-celery.

Previous PR: https://github.com/celery/django-celery-results/pull/56

### DIFF
--- a/django_celery_results/admin.py
+++ b/django_celery_results/admin.py
@@ -9,7 +9,7 @@ from django.utils.translation import ugettext_lazy as _
 try:
     ALLOW_EDITS = settings.DJANGO_CELERY_RESULTS['ALLOW_EDITS']
 except (AttributeError, KeyError):
-    ALLOW_EDITS = True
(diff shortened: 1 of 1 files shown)


maxmalysh:
As in PR #124, the only build failing is pydocstyle and it looks like the error is not related to my changes:

pydocstyle run-test: commands[0] | pydocstyle /home/travis/build/celery/django-celery-results/django_celery_results
254[4644] /home/travis/build/celery/django-celery-results$ /home/travis/build/celery/django-celery-results/.tox/pydocstyle/bin/pydocstyle django_celery_results
255django_celery_results/managers.py:96 in public method `store_result`:
256        D417: Missing argument descriptions in the docstring (argument(s) meta, traceback are missing descriptions in 'store_result' docstring)

PR #110, opened 2019-08-15
change double quotes to single quotes for support mysql5.6 query

lyr1994:
The origin query would have the following error when querying mysql5.6.
```
Unknown column 'tx_isolation' in 'where clause'
```

PR #135, opened 2020-02-27
Fix bug in warn_if_repeatable_read (introduced in PR #110).

arnau126:
In PR #110 some double quotes were overlooked, resulting in an invalid query.


lyr1994:
Thanks, my mistake.

arnau126:
Thanks for approving, @auvipy .
I would appreciate a quick release, as it is just a bug fix.

auvipy:
i will

PR #141, opened 2020-03-03
Bump supported django and python versions.

D3X:


kiyoya:
This looks like an easy fix. pydocstyle is complaining about missing docstring for some arguments while we actually separate them into two sections (Arguments vs Keyword Arguments).

Should we just merge those two sections everywhere?

kiyoya:
Oh, there is a PR for this: #139 (the test is still failing there though)

Issue #52, opened 2018-05-12
Chords not supported when using 'django-db' backend


asfaltboy:
I'm trying a simple chord example while using the 'django-db' result backend, and get back this error:

```
Traceback (most recent call last):
  File "/Users/pavel/dev/test_proj/src/batch/tests/test_batch.py", line 51, in test_stringing
    queue_batch()
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/local.py", line 191, in __call__
    return self._get_current_object()(*a, **kw)
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/app/task.py", line 380, in __call__
    return self.run(*args, **kwargs)
  File "/Users/pavel/dev/test_proj/src/batch/tasks.py", line 218, in queue_batch
    schedule_tasks(tasks, object)
  File "/Users/pavel/dev/test_proj/src/batch/tasks.py", line 196, in schedule_tasks
    chord(all_tasks_group)(set_object_state.s(object_id=object.id)).delay()
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/canvas.py", line 1196, in __call__
    return self.apply_async((), {'body': body} if body else {}, **options)
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/canvas.py", line 1237, in apply_async
    body=body, task_id=task_id, **options)
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/canvas.py", line 1251, in apply
    args=(tasks.apply(args, kwargs).get(propagate=propagate),),
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/result.py", line 635, in get
    on_message=on_message,
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/result.py", line 746, in join_native
    on_message, on_interval):
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/result.py", line 724, in iter_native
    on_message=on_message, on_interval=on_interval,
  File "/Users/pavel.virtualenvs/test_proj/lib/python3.6/site-packages/celery/backends/base.py", line 448, in iter_native
    return self.get_many(
AttributeError: 'DatabaseBackend' object has no attribute 'get_many'
```

Does anyone know if this is possible, and how can I work around this?

[...]

PR #144, opened 2020-05-12
Add chord syncronisation from the database

AllexVeldman:
This tracks chord synchronisation in the ChordCounter model instead of using the polling `unlock_chord` task.
In general it keeps a counter of the size of the chord header and decrements this counter on each return.
When the counter hits 0 the callback is triggered (borrowed from `unlock_chord`)

relates to #52


AllexVeldman:
Because the `select_for_update()` currently only has effect for postgresql, oracle and mysql this could potentially cause a race condition in the counter, should this behaviour be an option?

AllexVeldman:
Failing tests due to issues solved in #141 and #139

AllexVeldman:
@auvipy could you indicate if this is the correct approach?

auvipy:
i will look into this next week

ZviBaratz:
@auvipy @AllexVeldman Any updates on this?

AllexVeldman:
I haven't received any feedback yet, have been using a very similar implementation for a while now and have not yet run into any issues.

ZviBaratz:
Sounds good. @auvipy? 
Sorry for barging in here, just thought I'll check if I can do anything to help out. I :heart: Django + Celery.

auvipy:
ops damn I missed it for so long

auvipy:
> Sounds good. @auvipy?
> Sorry for barging in here, just thought I'll check if I can do anything to help out. I heart Django + Celery.

can you try this patch on your dev env?

ZviBaratz:
> can you try this patch on your dev env?

I tried running the tests but I'm getting a bunch of unrelated errors due to the `celery.five` import call in `models.py` (also discussed in #157). I'll try and figure out what's going on there but I'm somewhat new to celery so the implications of the transition from v. 4 to 5 are somewhat beyond me.

arnau126:
I've fixed this "unrelated errors" in PR: #163

auvipy (approved):

PR #145, opened 2020-05-29
Add Django 3.0 to test matrix

Palisand:


Palisand:
(Django 3.0 does not support Python 3.5)[https://docs.djangoproject.com/en/3.0/releases/3.0/]

PR #151, opened 2020-06-05
Add state argument to set method, to match changed signature in celery.

### LINKED ISSUE #150: Incompatibility with celery v4.4.4
Celery v4.4.4 [changed the signature](https://github.com/celery/celery/commit/0463bff0530b8aef8d31bd8039f245735295db59) of the `set()` method for results backends, which means that you currently get the following error when using celery 4.4.4:

```
set() takes 3 positional arguments but 4 were given
```

So we need to update that method here to accept the 4th argument.

solarissmoke:
Fixes #150.


solarissmoke:
Yeah, sorry, I will do that shortly.

solarissmoke:
@auvipy actually I don't think the tests are affected. The reason the build is currently failing is (a) because there were already a number of failures on master (see https://travis-ci.org/github/celery/django-celery-results/builds/692441736), and (b) because celery 4.4.4 is missing a dependency on `future`.

auvipy:
> @auvipy I don't think the tests are affected. The reason the build is currently failing is (a) because there were already a number of failures on master (see https://travis-ci.org/github/celery/django-celery-results/builds/692441736), and (b) because celery 4.4.4 is missing a dependency on `future`.
> 
> I don't think I will have time today to debug the pre-existing failures on master unfortunately.

ok

mchataigner:
I created another PR on celery project to fix this issue there.
https://github.com/celery/celery/pull/6173

mchataigner:
@auvipy I guess you may close this PR as well.

PR #161, opened 2020-10-02
Add support for GroupResult

### LINKED ISSUE #40: Cannot save GroupResult
I want to run multiple jobs as a group and want to store the result to watch the progress of the combined job.
According to the documentation I can use [GroupResult.save](http://docs.celeryproject.org/en/latest/reference/celery.result.html#celery.result.GroupResult.save) and restore it later.

When I try to do that I get the following error message:
```
ERROR	2017-11-10 09:05:26,293 exception 7586 139766003652352 Internal Server Error:
  File "venv/lib/python3.6/site-packages/django/core/handlers/exception.py", line 39, in inner
    response = get_response(request)
  File "venv/lib/python3.6/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "venv/lib/python3.6/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "venv/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "venv/lib/python3.6/site-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "venv/lib/python3.6/site-packages/rest_framework/views.py", line 489, in dispatch
    response = self.handle_exception(exc)
  File "venv/lib/python3.6/site-packages/rest_framework/views.py", line 449, in handle_exception
    self.raise_uncaught_exception(exc)
  File "venv/lib/python3.6/site-packages/rest_framework/views.py", line 486, in dispatch
    response = handler(request, *args, **kwargs)
  File "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", line 78, in post
    async_result.save()
  File "venv/lib/python3.6/site-packages/celery/result.py", line 833, in save
    return (backend or self.app.backend).save_group(self.id, self)
  File "venv/lib/python3.6/site-packages/celery/backends/base.py", line 386, in save_group
    return self._save_group(group_id, result)
AttributeError: 'DatabaseBackend' object has no attribute '_save_group'
```

This makes sense as `django_celery_results.backends.database.DatabaseBackend` does not implement this method. `celery.backends.database.DatabaseBackend` Implements it but it is not used when set `CELERY_RESULT_BACKEND = 'django-db'`.

Did I configure this incorrectly or is this simply not implemented?

What I want to do is return the task_id of the group when a combined job is submitted and then poll the status of the group to see when all tasks are finished and not check every sub-task individually.

AlecTroemel:
closes #40 

Adds `GroupResult` model, group result model manager, and implements the `group` methods on the `DatabaseBackend`. this PR still needs tests, and probably some other tweaks. but its in a "working" state so I thought it would be ready for more eyes


auvipy (changes requested):
needs rebase

django_celery_results/urls.py
```diff
@@ -24,4 +24,14 @@
         views.task_status,
         name='celery-task_status'
     ),
+    url(
```
auvipy, line 27: better use path()

auvipy (changes requested):
also, check the Travis failures.

auvipy (changes requested):
also please rebase and let us know if it is ready for further review

django_celery_results/urls.py
AlecTroemel: @auvipy would you prefer that I just use path for the GroupResult views, or switch to `path` for the AsyncTask views as well?

AlecTroemel:
@auvipy I've rebased, but I'm not sure what to make of the pipeline failures. it seems like the root of it is a test trying to call `get` for a `ChordCounter` that doesn't exist

```
 django_celery_results.models.ChordCounter.DoesNotExist: ChordCounter matching query does not exist.
```

But this MR does not (or at least should not) make any changes to the `ChordCounter` related code :thinking: . I'm not seeing any file changes in this MR related to the ChordCounter tests

arnau126:
@AlecTroemel these `ChordCounter` are not this PR fault.
I've fixed them in #165  (but not merged yet).

Ixiodor:
I'm starting to get

> django_celery_results.models.ChordCounter.DoesNotExist: ChordCounter matching query does not exist.

andi don't know why.

arnau126:
@Ixiodor , could you provide the whole traceback, and the `celery` and `django-celery-results` versions you're using?

django_celery_results/migrations/0009_groupresult.py
```diff
@@ -0,0 +1,30 @@
...
+    dependencies = [
+        ('django_celery_results', '0008_chordcounter'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='GroupResult',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('group_id', models.CharField(db_index=True, help_text='Celery ID for the Group that was run', max_length=191, unique=True, verbose_name='Group ID')),
```
AllexVeldman, line 17: Make sure to use 
```
max_length=getattr(
                        settings,
                        'DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH',
                        255
                    ),
```
to respect the settings.
see https://github.com/celery/django-celery-results/pull/172

auvipy:
any update?

Ixiodor:
> Traceback (most recent call last):
>   File "/home/ubuntu/.local/lib/python3.7/site-packages/celery/app/trace.py", line 477, in trace_task
>     uuid, retval, task_request, publish_result,
>   File "/home/ubuntu/.local/lib/python3.7/site-packages/celery/backends/base.py", line 156, in mark_as_done
>     self.on_chord_part_return(request, state, result)
>   File "/home/ubuntu/.local/lib/python3.7/site-packages/django_celery_results/backends/database.py", line 113, in on_chord_part_return
>     .get(group_id=gid)
>   File "/home/ubuntu/.local/lib/python3.7/site-packages/django/db/models/query.py", line 431, in get
>     self.model._meta.object_name
> django_celery_results.models.ChordCounter.DoesNotExist: ChordCounter matching query does not exist.
> 

My stack

auvipy (changes requested):
you need to rebase this on top of master

django_celery_results/urls.py
auvipy: all

fmoralesalcaide:
Any update on this? Thank you

Guanyan1996:
wait this update to master .

fmoralesalcaide:
I can see that this last CI tests fail all with the same error:

image

Could you re-check this? Thank you very much, I hope this gets merged soon.

AlecTroemel:
hey all, sorry for being MIA for so long. I've brought this branch up to date with master, we'll see if those odd test errors magically go away

I'll try my best to be active on this MR, hopefully we can get this merged!

auvipy:
will check & review this soon

django_celery_results/models.py
```diff
@@ -137,15 +138,71 @@ class ChordCounter(models.Model):
...
         """
         return GroupResult(
             self.group_id,
             [result_from_tuple(r, app=app)
              for r in json.loads(self.sub_tasks)],
-            app=app,
         )
+
+
+class GroupResult(models.Model):
```
AllexVeldman, line 152: This name clashes with `from celery.result import GroupResult`

django_celery_results/backends/database.py
```diff
@@ -115,14 +117,39 @@ def _forget(self, task_id):
...
+    def _delete_group(self, group_id):
+        try:
+            self.GroupModel._default_manager.get_group(group_id).delete()
+        except self.TaskModel.DoesNotExist:
+            pass
 
     def apply_chord(self, header_result, body, **kwargs):
         """Add a ChordCounter with the expected number of results"""
         results = [r.as_tuple() for r in header_result]
         data = json.dumps(results)
```
AllexVeldman, line 151: Once this PR is done we could use it in storing the head group for chords instead of json encoding the `header_result` here and decoding it again in `on_chord_part_return`

django_celery_results/models.py
AllexVeldman, line 152: This name clashes with `from celery.result import GroupResult`

AllexVeldman (changes requested):

django_celery_results/models.py
AlecTroemel: I agree that it does.. however [in the celery docs](https://docs.celeryproject.org/en/latest/faq.html?highlight=TaskResult#what-purpose-does-the-database-tables-created-by-django-celery-results-have) they refer to a GroupResult model in this project, even though it doesnt exist yet (maybe it did at some point?)

[this fork](https://github.com/celery/django-celery-results/compare/master...CantemoInternal:cantemo) of django-celery-results used `TaskSetMeta` for the model name. It seems like `celery.result.GroupResult` used to be called `TaskSetResult`, so there was probably a collision there as well.


if the concern is with collisions specifically in the `models.py` file, we could use the full path `celery.result.GroupResult` where the actual celery class is needed.  But maybe its better to avoid collisions altogether

AlecTroemel: I've switched to more explicit imports in `models.py` so at least we dont have name clashes in the file

django_celery_results/urls.py
AlecTroemel: I've converted all the url's to path. One side effect of this is that I had to create a custom path converter, since the regex for task/group id's is a little different then the built in converters

django_celery_results/urls.py
```diff
@@ -6,21 +6,49 @@
     URL  to :func:`~celery.views.task_status`.
 """
 
-from django.conf.urls import url
+from django.conf.urls import path, register_converter
```
auvipy, line 9: i was thinking about importing from django.urls as we support 2.2+ in master

django_celery_results/urls.py
```diff
@@ -6,21 +6,49 @@
...
 
 from . import views
 
-task_pattern = r'(?P<task_id>[\w\d\-\.]+)'
+
+class TaskPatternConverter:
```
auvipy, line 14: so  we still need a custom converter?

django_celery_results/managers.py
```diff
@@ -57,7 +57,44 @@ def _inner(*args, **kwargs):
     return _outer
 
 
-class TaskResultManager(models.Manager):
+class ResultManager(models.Manager):
```
auvipy, line 60: this is not a backward-incompatible change, right?

django_celery_results/urls.py
AlecTroemel: I believe so. Django has a `uuid` converter, but the pattern we're using is a little different. Specifically, our pattern allows for `.`, while uuid does not. For what its worth I've never seen a celery task ID with a period, but I haven't found the official docs on that

django_celery_results/urls.py
AlecTroemel: good with me!

django_celery_results/managers.py
AlecTroemel: right, I believe it is compatible. 

The git diff looks a little goofy, but what I've done is pulled out the common logic between task & group managers out into a parent class. `TaskResultManager` should act identically to how it does now in master

auvipy: make sense

auvipy:
it would be great if you also pull changes from master branch

AlecTroemel:
i want to address that LGTM error concerning imports. To avoid clobbering the namespace with 2 things named `GroupResult`, Im currently just importing celery and using the full path to that instance of `GroupResult`. The issue is that the `celery.states` module only works if I also import that :facepalm: . 

One thought is to use an `import as` for the vanilla celery `GroupResult`.. something like this

```python
# in models.py
from celery.result import GroupResult as CeleryGroupResult
```

then all the celery imports together would look like this 
```python
# in models.py
from celery import states
from celery.result import result_from_tuple
from celery.result import GroupResult as CeleryGroupResult
```

thoughts?

fmoralesalcaide:
I really appreciate the effort put into this PR. Thank you guys.

auvipy:
@arnau126 @asfaltboy please have a look when you have time. need more eyes before merge

vesnikos:
this looks good for v2.1 ?

auvipy:
need some more time to review & get a look from some other guys

thedrow:
@AlecTroemel If this is still WIP, please convert this PR to a draft.

AlecTroemel:
@thedrow nope this PR is ready to go, barring any feedback. thanks for changing the title!

django_celery_results/models.py
```diff
@@ -143,9 +144,71 @@ def group_result(self, app=None):
             app (Celery): app instance to create the GroupResult with.
 
         """
-        return GroupResult(
+        return CeleryGroupResult(
             self.group_id,
             [result_from_tuple(r, app=app)
              for r in json.loads(self.sub_tasks)],
-            app=app,
```
auvipy, line 150: can you make sure removing this won't create any regression or backward compat issue?

django_celery_results/models.py
```diff
@@ -143,9 +144,71 @@ def group_result(self, app=None):
...
+    """Task Group result/status."""
+
+    group_id = models.CharField(
+        max_length=getattr(
+            settings,
+            "DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH",
+            255
+        ),
+        unique=True,
+        db_index=True,
```
auvipy, line 164: instead of using db_index in the field isn't it better to use the model meta Index option and more future proof?

auvipy (changes requested):

Guanyan1996:
how long time or version can I use this changed?

django_celery_results/models.py
AlecTroemel: Im honestly not sure why I removed this line, we still allow the app to be passed in as a kwarg. I'm going to add it back

django_celery_results/models.py
AlecTroemel: Reading your comment more closely I've honestly never heard of the `model meta index option`.. I cant find anything in the django docs, you have a reference you could point me towards? 

I copied `db_index`  over from the "shared"  fields with the `TaskResult` model. If you're suggesting a change to that model as well would that be better as a separate PR?

vesnikos: > instead of using db_index in the field isn't it better to use the model meta Index option and more future proof?

Not sure if it's needed. Defining the index in the meta part, is more for defining a different more exotic type of index (ie for postgres: https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/indexes/) ,  or a partial index rather keeping an ol' trusty b-tree index.  https://docs.djangoproject.com/en/3.2/ref/models/options/#django.db.models.Options.indexes
 
Actually, you could be even less verbose by omitting the db_index=True altogether as 
> Note that when unique is True, you don’t need to specify db_index, because unique implies the creation of an index. [src](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.Field.unique)

AlecTroemel: interesting! I'm good with removing `db_index=True` if its redundant

AlecTroemel: @auvipy thoughts?

auvipy: db_index is now implemented as model meta like this https://docs.djangoproject.com/en/3.2/ref/models/options/#django.db.models.Options.indexes instead of defining in the filed they can be defined in the model meta option

Guanyan1996:
Hello,bro....When will this feature plan end? It has been waiting for half a year.thx
@auvipy @AlecTroemel

django_celery_results/models.py
```diff
@@ -136,6 +144,13 @@ class ChordCounter(models.Model):
         )
     )
 
+    class Meta:
```
auvipy, line 147: curious why two class meta?

django_celery_results/migrations/0010_auto_20210416_1502.py
```diff
@@ -0,0 +1,103 @@
+# Generated by Django 3.2 on 2021-04-16 15:02
```
auvipy, line 1: remove this migration mile as well and freshly generate with the new changes in models

django_celery_results/urls.py
```diff
@@ -6,21 +6,49 @@
...
-        r'^%s/done/?$' % task_pattern,
+    path(
+        r'<task_pattern:task_id>/done/',
         views.is_task_successful,
         name='celery-is_task_successful'
     ),
-    url(
-        r'^%s/status/?$' % task_pattern,
+    path(
+        r'<task_pattern:task_id>/status/',
```
auvipy, line 40: there should not be r prefix

auvipy (changes requested):

django_celery_results/models.py
AlecTroemel: oops, thats a typo!

AlecTroemel: I thought I understood.. but I dont see the second "class meta" you're referring to

django_celery_results/urls.py
AllexVeldman: @AlecTroemel the task id is provided by `kombu.uuid()` which is an alias for the [builtin uuid4](https://docs.python.org/dev/library/uuid.html#uuid.uuid4).
[according to wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format) it will only contain HEX characters and hyphens (`[0-1a-zA-Z\-]`) So I think the custom converter is not needed if it's only used for the UUID pattern

django_celery_results/models.py
auvipy: i was my mistake :D

auvipy:
> Hello,bro....When will this feature plan end? It has been waiting for half a year.thx
> @auvipy @AlecTroemel

some reviews were needed to be addressed but expect very soon

django_celery_results/urls.py
AlecTroemel: @AllexVeldman I don't disagree with you, however the old task pattern does allow for periods `.`.  Again I've never actually seen a period in a task ID, so it could be that there never have been and  never will be any. If that's the case I'd love to use the default uuid converter! 

I'll do some digging in the commit history to see if I can learn anything about the context of the old pattern

AlecTroemel: hmm, not a whole lot of history with this file https://github.com/celery/django-celery-results/commits/master/django_celery_results/urls.py

django_celery_results/backends/database.py
```diff
@@ -155,22 +197,18 @@ def on_chord_part_return(self, request, state, result, **kwargs):
             if deps.ready():
                 callback = maybe_signature(request.chord, app=self.app)
                 trigger_callback(
-                    app=self.app,
```
auvipy, line 158: revert this unrelated change.

auvipy:
also what about docs updates?

AlecTroemel:
I've  updated the `docs/introduction.txt` text, not sure if there are other places needing updates. Should I update the `Changelog` now? Or do you want to wait until a new version is released?

devang-aubergine:
Any updates on the same? @AlecTroemel  @auvipy

auvipy (changes requested):
can you pull changes from master?

auvipy (approved):

auvipy:
thanks a lot. going to release this soon. please help us if any issues are reported after release.

auvipy:
test failing after push

=================================== FAILURES ===================================
___________ test_DatabaseBackend.test_apply_chord_header_result_arg ____________

self = <t.unit.backends.test_database.test_DatabaseBackend object at 0x7f1119ef8890>

    def test_apply_chord_header_result_arg(self):
        """Test if apply_chord can handle Celery <= 5.1 call signature"""
        gid = uuid()
        tid1 = uuid()
        tid2 = uuid()
        # Celery < 5.1
>       self.b.apply_chord(group, self.add.s())

t/unit/backends/test_database.py:700: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
django_celery_results/backends/database.py:165: in apply_chord
    header_result = self.app.GroupResult(*header_result_args)
.tox/py-django32/lib/python3.7/site-packages/celery/result.py:886: in __init__
    ResultSet.__init__(self, results, **kwargs)
.tox/py-django32/lib/python3.7/site-packages/celery/result.py:539: in __init__
    self._on_full = ready_barrier or barrier(results)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <vine.synchronization.barrier object at 0x7f111a1896d0>
promises = <AsyncResult: 24c11db6-ffe0-4008-997d-7e6e1c02101d>, args = None
kwargs = None, callback = None, size = None

    def __init__(self, promises=None, args=None, kwargs=None,
                 callback=None, size=None):
        self.p = promise()
        self.args = args or ()
        self.kwargs = kwargs or {}
        self._value = 0
        self.size = size or 0
        if not self.size and promises:
            # iter(l) calls len(l) so generator wrappers
            # can only return NotImplemented in the case the
            # generator is not fully consumed yet.
>           plen = promises.__len__()
E           AttributeError: 'AsyncResult' object has no attribute '__len__'

auvipy:
have o delay the release

AllexVeldman:
@AlecTroemel I noticed that `GroupResult.restore(gid)` returns `List[AsyncResult]` instead of `GroupResult`. is this correct?

AllexVeldman:
I also think we might lose nested groups in the current implementation due to 
```python
content_type, content_encoding, result = self.encode_content(
            [r.id for r in group_result]
        )
```

Using 
```python
content_type, content_encoding, result = self.encode_content(
            group_result.as_tuple()
        )

res["result"] = result_from_tuple(decoded_result, app=self.app)
```

Would ensure the restored group equals the stored group

AllexVeldman:
https://github.com/AllexVeldman/django-celery-results/blob/bugfix/nested-groups/django_celery_results/backends/database.py contains a WIP POC.

auvipy:
would be great if you open a draft PR

AllexVeldman:
> would be great if you open a draft PR

https://github.com/celery/django-celery-results/pull/204

auvipy:
yup i saw that, thanks again

Issue #167, opened 2020-10-21
Invalid migration value for `max_length` ChordCounter.group_id on master


tomwojcik:
I use this lib from commit ed94080cea3d68a38f9ff048657a37b08e74fc35 with fixes for Celery 5.x 

After making migrations it adds.

```
Migrations for 'django_celery_results':
  /usr/local/lib/python3.8/site-packages/django_celery_results/migrations/0009_auto_20201021_1026.py
    - Alter field group_id on chordcounter
```

https://github.com/celery/django-celery-results/commit/ed94080cea3d68a38f9ff048657a37b08e74fc35#diff-fb5af8c04137d0318c8f7a50ddb6697523e311bf7773374ff886817027a657e6R116-R125

vs

https://github.com/celery/django-celery-results/commit/ed94080cea3d68a38f9ff048657a37b08e74fc35#diff-e1f276feafdcddf6dc83ab6214d028c3d98008407e4320bd479d6c69d427dd48R22-R27

tomwojcik:
I can't link PR so here it is https://github.com/celery/django-celery-results/pull/168

tomwojcik:
https://github.com/celery/django-celery-results/pull/169

PR #168, opened 2020-10-21
fix default max_length in migration of ChordCounter.group_id

tomwojcik:


tomwojcik:
https://github.com/celery/django-celery-results/issues/167

PR #182, opened 2021-01-16
Add task_args and task_kwargs to searchable fields

gopackgo90:
Adding `task_args` and `task_kwargs` to searchable fields in the admin. This is particularly useful in case the arguments provided are specific ids of objects or unique fields.


gopackgo90:
@auvipy, if you're doing a release soon, do you think this could be added to it? Thanks!

auvipy (approved):

PR #186, opened 2021-01-27
Don't raise an error when ChordCounter is not found

### LINKED ISSUE #177: ChordCounter DoesNotExist
I get this error for each task i launch:
```
  File "/home/ubuntu/venvs/earthalytics-api/lib/python3.8/site-packages/celery/app/trace.py", line 476, in trace_task
    mark_as_done(
  File "/home/ubuntu/venvs/earthalytics-api/lib/python3.8/site-packages/celery/backends/base.py", line 156, in mark_as_done
    self.on_chord_part_return(request, state, result)
  File "/home/ubuntu/venvs/earthalytics-api/lib/python3.8/site-packages/django_celery_results/backends/database.py", line 112, in on_chord_part_return
    ChordCounter.objects.select_for_update()
  File "/home/ubuntu/venvs/earthalytics-api/lib/python3.8/site-packages/django/db/models/query.py", line 429, in get
    raise self.model.DoesNotExist(
django_celery_results.models.ChordCounter.DoesNotExist: ChordCounter matching query does not exist.
```
Reinstalling all seems to fix for a while, after start that again.
Celery 5.0.2 and 2.0.0 django_results

AllexVeldman:
When upgrading from <2.0.0 while current chords are not finished, the ChordCounter object does not exist, breaking the chord.
The `celery.chord_unlock` Task created by the legacy chord should unlock any existing chord while new chord will create the proper ChordCounter object

closes #177

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -137,8 +137,11 @@ class DatabaseBackend(BaseDictBackend):
             # SELECT FOR UPDATE is not supported on all databases
             chord_counter = (
                 ChordCounter.objects.select_for_update()
-                .get(group_id=gid)
+                .filter(group_id=gid).first()
             )
+            if chord_counter is None:
+                logger.warning("Can't find ChordCounter for Group %s", gid)
+                return
             chord_counter.count -= 1
             if chord_counter.count != 0:
                 chord_counter.save()
--- a/t/unit/backends/test_database.py
+++ b/t/unit/backends/test_database.py
@@ -725,6 +725,17 @@ class test_DatabaseBackend:
 
         request.chord.delay.assert_called_once()
 
+    def test_on_chord_part_return_counter_not_found(self):
+        """Test if the chord does not raise an error if the ChordCounter is not found
+
+        Basically this covers the case where a chord was created with a version
+        <2.0.0 and the update was done before the chord was finished
+        """
+        request = mock.MagicMock()
+        request.id = uuid()
+        request.group = uuid()
+        self.b.on_chord_part_return(request=request, state=None, result=None)
+
     def test_callback_failure(self):
         """Test if a failure in the chord callback is properly handled"""
         gid = uuid()


Rjevski:
I tested this (related to the other issue), it _seems_ to work fine regarding the steps to reproduce I posted on the issue. Now I get a `WARNING/ForkPoolWorker-1] Can't find ChordCounter for Group 8c429463-16c5-4844-b412-258578341ce5` but it seems to continue fine.

I'm not familiar with writing Celery result backends, but what actually calls `on_chord_part_return`? What puzzles me is that the PR that introduced this issue merely _added_ functions and doesn't seem to have anything calling them, so I'm not sure what that function is supposed to do and what potential impacts (that we don't see yet) it will have if the function just returns?

If it helps I can upload my test project in a repo so you can easily test this on your side. You'll need a Postgres DB (to be as close to the production use case I have, though now that I think of it I don't think it has any impact) and a Redis server.

auvipy:
> When upgrading from <2.0.0 while current chords are not finished, the ChordCounter object does not exist, breaking the chord.
> The `celery.chord_unlock` Task created by the legacy chord should unlock any existing chord while new chord will create the proper ChordCounter object
> 
> closes #177

I'm not familiar with writing Celery result backends, but what actually calls on_chord_part_return? What puzzles me is that the PR that introduced this issue merely added functions and doesn't seem to have anything calling them, so I'm not sure what that function is supposed to do and what potential impacts (that we don't see yet) it will have if the function just returns?

-- any thoughts about addressing this?

AllexVeldman:
> I tested this (related to the other issue), it _seems_ to work fine regarding the steps to reproduce I posted on the issue. Now I get a `WARNING/ForkPoolWorker-1] Can't find ChordCounter for Group 8c429463-16c5-4844-b412-258578341ce5` but it seems to continue fine.

I added that warning to at least indicate something happend that in normal operation shouldn't.

> I'm not familiar with writing Celery result backends, but what actually calls `on_chord_part_return`? What puzzles me is that the PR that introduced this issue merely _added_ functions and doesn't seem to have anything calling them, so I'm not sure what that function is supposed to do and what potential impacts (that we don't see yet) it will have if the function just returns?

`on_chord_part_return` is called in `mark_as_done`, `mark_as_failure` and `mark_as_revoked` on the base Backend class https://github.com/celery/celery/blob/29eda054555fa95c83210e5e6bc3e839c80bcd3b/celery/backends/base.py#L156
The default implementation is a no-op: https://github.com/celery/celery/blob/29eda054555fa95c83210e5e6bc3e839c80bcd3b/celery/backends/base.py#L590.

This is expected as the base backend relies on `celery.chord_unlock` task to unlock Chords (like `django_celery_results < 2.0.0`).

In my original implementation I let the exception raise because not finding the ChordCounter object would mean your callback will never be queued so it's at least clear something went wrong.
To allow this upgrade path of having a non-finished Chord while upgrading I don't see another viable option than to default back to the base Backend implementation (the no-op).

@auvipy what do you think?

auvipy:
> > I tested this (related to the other issue), it _seems_ to work fine regarding the steps to reproduce I posted on the issue. Now I get a `WARNING/ForkPoolWorker-1] Can't find ChordCounter for Group 8c429463-16c5-4844-b412-258578341ce5` but it seems to continue fine.
> 
> I added that warning to at least indicate something happend that in normal operation shouldn't.
> 
> > I'm not familiar with writing Celery result backends, but what actually calls `on_chord_part_return`? What puzzles me is that the PR that introduced this issue merely _added_ functions and doesn't seem to have anything calling them, so I'm not sure what that function is supposed to do and what potential impacts (that we don't see yet) it will have if the function just returns?
> 
> `on_chord_part_return` is called in `mark_as_done`, `mark_as_failure` and `mark_as_revoked` on the base Backend class https://github.com/celery/celery/blob/29eda054555fa95c83210e5e6bc3e839c80bcd3b/celery/backends/base.py#L156
> The default implementation is a no-op: https://github.com/celery/celery/blob/29eda054555fa95c83210e5e6bc3e839c80bcd3b/celery/backends/base.py#L590.
> 
> This is expected as the base backend relies on `celery.chord_unlock` task to unlock Chords (like `django_celery_results < 2.0.0`).
> 
> In my original implementation I let the exception raise because not finding the ChordCounter object would mean your callback will never be queued so it's at least clear something went wrong.
> To allow this upgrade path of having a non-finished Chord while upgrading I don't see another viable option than to default back to the base Backend implementation (the no-op).
> 
> @auvipy what do you think?

thats seems a reasonable default. I would merge this PR until we have any better solution

auvipy:
> I'm not familiar with writing Celery result backends, but what actually calls `on_chord_part_return`? What puzzles me is that the PR that introduced this issue merely _added_ functions and doesn't seem to have anything calling them, so I'm not sure what that function is supposed to do and what potential impacts (that we don't see yet) it will have if the function just returns?

the most helpful thing will be to digging the code & trying to figure out a better solution for the thing you are thinking could be improved :)

auvipy:
@arnau126 @asfaltboy if you don't have any other improvement offer I would like to merge this and release it as stopgap solution

Rjevski:
I think this solution is good. I've tried it locally and it appears to work fine - originally I was debugging a much larger issue (which involved the DB filling up with a large amount of bogus TaskResults - until it eventually ran out of space) but I assume this is related to my own code as I couldn't replicate it with the example code I provided (even adjusting it to match the chord pattern I have in the real codebase).

auvipy (approved):

auvipy:
thanks all!

PR #200, opened 2021-05-25
Add GitHub workflow for unittesting

AllexVeldman:
Run the unittests in GitHub Actions.
This is a possible replacement for the current TravisCI setup.

The unittests rely on a Postgres Docker service so `HOST` and `PASSWORD` had to be added to `settings.py`.
As I'm not familiar with the release process I'm not 100% sure this does not impact releasing.

I'm also not sure how coverage is setup currently so that would have to be added by one of the maintainers.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,71 @@
+name: Test
+
+on:
+    pull_request:
+        branches: [ "master" ]
+
+    push:
+        branches: ["master"]
+
+jobs:
+    matrix:
+        name: Unittest Matrix
+        runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


auvipy:
thanks for handling this, was wondering why the actions are not triggered yer

AllexVeldman:
> thanks for handling this, was wondering why the actions are not triggered yer

Because the master branch does not contain the workflow yet (they are triggered only on push or PR to master)
https://github.com/AllexVeldman/django-celery-results/actions/runs/874784039 is a run of the workflow on push while i was working on it

auvipy (approved):

PR #204, opened 2021-06-01
store GroupResult as_tuple()

AllexVeldman:
Store the entire group as serialised by `as_tuple()` instead of the list of Task id's.

This will also change the return value for `_restore_group()["result"]` from `List[AsyncResult]` to `GroupResult` ensuring 
```python
group = GroupResult(group_id)
group.save()
group == <backend>.restore_group(group_id=group_id)
```

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -4,7 +4,7 @@ import json
 from celery import maybe_signature
 from celery.backends.base import BaseDictBackend
 from celery.exceptions import ChordError
-from celery.result import allow_join_result, GroupResult
+from celery.result import allow_join_result, result_from_tuple, GroupResult
 from celery.utils.serialization import b64encode, b64decode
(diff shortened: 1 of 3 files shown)


auvipy:
should we wait for this or cut a new release from master?

AllexVeldman:
@auvipy I don't have the time this week to properly test this and compare it with the `GroupResult.restore()` from the redis backend. As the restore functionality is new, this feature might just be broken if you release from master now. it will unblock upgrading to celery 5.1.0 so I think you should go for it.

auvipy:
yes possible, i will check

AllexVeldman:
@auvipy I think this should be good to go, I reverted the change in the ChordCounter for now to minimize the scope to just the bugfix

auvipy (approved):

auvipy:
I will try to release a new version today

Issue #206, opened 2021-06-15
2.1.0 migration hardcodes max_length for group_id and task_id fields


AllexVeldman:
with https://github.com/celery/django-celery-results/blob/master/django_celery_results/migrations/0009_groupresult.py
the `max_length` for `task_id` and `group_id` fields on `TaskResult`, `ChordCounter` and `GroupResult` are all hardcoded to 191.

This should have been taken from the `DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH` setting.

```python
max_length=getattr(
    settings,
    'DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH',
    255
),
```

AllexVeldman:
@auvipy @arnau126 I think we should edit the migration in-place as this can potentially break existing rows in the database when applied.

PR #207, opened 2021-06-15
Use the DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH for group_id/task_id

AllexVeldman:
closes https://github.com/celery/django-celery-results/issues/206


auvipy (approved):

vanschelven:
I see how this is a solution for the case where `DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH` is not set... but I wonder about the following scenario:

* Install `django-celery-results`
* Run migrations
* Tune `DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH`
* Either your database won't be updated, or you'll have the same complaint from #206 about missing migrations.

My take on this is: using settings to affect your models is always going to give trouble. Better have the model just be 255 long, and use the setting at application-layer exclusively (i.e. not in the DB).

AllexVeldman:
I agree, from what I can gather it is specifically added for MySQL databases.
never needed to touch this setting myself so I just followed how the migration were set up before this broke.

PR #215, opened 2021-06-28
fix group status view return data, add tests for it

### LINKED ISSUE #214: <id>/group/status/ is broken
It seems that this part of code was not tested at all.

Error:

```python
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/local/lib/python3.9/site-packages/django_celery_results/views.py", line 49, in group_status
    retval = result.result
```

Possible reason (will not test, will remove current master as very dirty), but:

https://github.com/celery/django-celery-results/blob/3ddad5b72eeda38de8738cc6cc13f54571cc87ae/django_celery_results/views.py#L49

Not clear what expected to receive as no documentation for this part, but probably it should be
```
retval = result.results
```

With additional `s`, as it is group.

AlecTroemel:
closes #214 

api view will return json of this format

```json
{
    "group": {
        "id": "ID_HERE",
        "results": [
            {"result": "result here or null", "status": "STATUS"}
        ]
    }
}
```

### DIFF
--- a/django_celery_results/views.py
+++ b/django_celery_results/views.py
@@ -51,5 +51,4 @@ def group_status(request, group_id):
         for async_result in result.results
     ]
     response_data = {'id': group_id, 'results': retval}
-    print(response_data)
     return JsonResponse({'group': response_data})
--- a/django_celery_results/views.py
+++ b/django_celery_results/views.py
@@ -44,8 +44,12 @@ def is_group_successful(request, group_id):
 
 
 def group_status(request, group_id):
-    """Return task status and result in JSON format."""
+    """Return group id and its list of async results status and result in JSON format."""
     result = GroupResult.restore(group_id)
-    retval = result.result
-    response_data = {'id': group_id, 'result': retval}
-    return JsonResponse({'task': response_data})
+    retval = [
+        {"result": async_result.result, "status": async_result.status}
+        for async_result in result.results
+    ]
+    response_data = {'id': group_id, 'results': retval}
+    print(response_data)
+    return JsonResponse({'group': response_data})
--- a/django_celery_results/views.py
+++ b/django_celery_results/views.py
@@ -44,7 +44,7 @@ def is_group_successful(request, group_id):
 
 
 def group_status(request, group_id):
-    """Return group id and its list of async results status and result in JSON format."""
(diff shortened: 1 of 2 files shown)


django_celery_results/views.py
```diff
@@ -44,8 +44,12 @@ def is_group_successful(request, group_id):
...
     result = GroupResult.restore(group_id)
-    retval = result.result
-    response_data = {'id': group_id, 'result': retval}
-    return JsonResponse({'task': response_data})
+    retval = [
+        {"result": async_result.result, "status": async_result.status}
+        for async_result in result.results
+    ]
+    response_data = {'id': group_id, 'results': retval}
+    print(response_data)
```
auvipy, line 54: print(response_data) isnt needed i think

django_celery_results/views.py
```diff
@@ -44,8 +44,12 @@ def is_group_successful(request, group_id):
...
-    """Return task status and result in JSON format."""
+    """Return group id and its async results status & result in JSON format."""
     result = GroupResult.restore(group_id)
-    retval = result.result
-    response_data = {'id': group_id, 'result': retval}
-    return JsonResponse({'task': response_data})
+    retval = [
+        {"result": async_result.result, "status": async_result.status}
+        for async_result in result.results
+    ]
```
AllexVeldman, line 52: Consider using `GroupResult.as_tuple()` to also return nested groups.

This has the added benefit of mainly using the GroupResult API to serialise the GroupResult instead of reimplementing it.

AlecTroemel: hmm. It looks like `as_tuple` doesn't include the status of the tasks, so im not sure we'll be able to use it here. I hadn't thought abut nested groups though, good catch.

AllexVeldman: Ah right it just contains the IDs

AllexVeldman (approved):

auvipy (approved):

PR #235, opened 2021-09-27
feat: raw delete expired instead of `Queryset.delete`

### LINKED ISSUE #232: DatabaseBackend.cleanup causes OOM kill for large expired result
# Problem
`DatabaeBackend.cleanup()` method uses Django's `QuerySet.delete` method - which could be stupidly slow and cause a LOT of memory bloting since this operation loads all object in-memory before deleting. Following is the quote explaining it from the [docs](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#delete)

> The delete() method does a bulk delete and does not call any delete() methods on your models. It does, however, emit the pre_delete and post_delete signals for all deleted objects (including cascaded deletions).
>
>Django needs to fetch objects into memory to send signals and handle cascades. However, if there are no cascades and no signals, then Django may take a fast-path and delete objects without fetching into memory. For large deletes this can result in significantly reduced memory usage. The amount of executed queries can be reduced, too.
>
>ForeignKeys which are set to on_delete DO_NOTHING do not prevent taking the fast-path in deletion.
>
>Note that the queries generated in object deletion is an implementation detail subject to change.

# Possible Solution
1. Bypassing Django's ORM and singals layer with either or the following:
  a. using raw SQL for deletion - with `RawQuerySet` to delete the selected raws (Note: that like `QuerySet` they are lazy too, so would need to make sure the query is fired)
  b. using raw SQL for deletion - with `connection.cursor.execute` API
  c. using `QuerySet._raw_delete(using)` abstraction - only issue being that the method is private

# Other notes/reads
- [Ticker-9519](https://code.djangoproject.com/ticket/9519) - discussing adding some `bulk_delete` method which is more efficient with deleting large volume of record

# To conclude
Since celery-result could grow very fast (for system that executes lots of tiny celery tasks) - for such system this behaviour could cause OOM kills and the actual "delete/cleanup operation" is never performed. 

PS: I would like to contribute this fix, if given green signal here.

daadu:
This is avoid unncessary loading of objects in-memory in certain conditions

Closes #232

### DIFF
--- a/django_celery_results/managers.py
+++ b/django_celery_results/managers.py
@@ -8,7 +8,7 @@ from celery.utils.time import maybe_timedelta
 from django.conf import settings
 from django.db import connections, models, router, transaction
 
-from .utils import now
+from .utils import now, raw_delete
 
 W_ISOLATION_REP = """
 Polling results with transaction isolation level 'repeatable-read'
@@ -88,7 +88,7 @@ class ResultManager(models.Manager):
     def delete_expired(self, expires):
         """Delete all expired results."""
         with transaction.atomic():
-            self.get_all_expired(expires).delete()
+            raw_delete(queryset=self.get_all_expired(expires))
 
 
 class TaskResultManager(ResultManager):
--- a/django_celery_results/utils.py
+++ b/django_celery_results/utils.py
@@ -15,3 +15,11 @@ def now():
         return now_localtime(timezone.now())
     else:
         return timezone.now()
+
+
+def raw_delete(queryset):
+    """
+    Raw delete given queryset. This is to avoid loading objects in-memory (in certain conditions
+    like - singals or cascade check) while deleting large number of rows.
+    """
+    return queryset._raw_delete(queryset.db)


daadu:
There is already a test that checks if the `delete_expired` method is functioning properly - so in case of breaking change in "protected Django method" - we will be able to catch it (Hopefully!)

daadu:
@auvipy  - Requesting review

auvipy:
thanks for the pr

auvipy (changes requested):
I still want to see some new unit tests

daadu:
@auvipy unit test  against `raw_delete` added as suggested.

django_celery_results/managers.py
```diff
@@ -88,7 +88,7 @@ def get_all_expired(self, expires):
     def delete_expired(self, expires):
         """Delete all expired results."""
         with transaction.atomic():
-            self.get_all_expired(expires).delete()
+            raw_delete(queryset=self.get_all_expired(expires))
```
auvipy, line 91: another thing i want to be assured that running makemigrations detect any change in migrations

django_celery_results/managers.py
```diff
@@ -88,7 +88,7 @@ def get_all_expired(self, expires):
     def delete_expired(self, expires):
```
auvipy, line 88: tests for this method needs to be updated as well

auvipy (changes requested):

django_celery_results/managers.py
daadu: I added [similar test for `django-silk` two days back](https://github.com/jazzband/django-silk/pull/495/files), where i simply called management command `makemigrations silk --check --dry-run` - if there are pending migrations then it returns with exit code 1.

Should we implement the same?

auvipy: that would be great!

daadu: There is already this test which does that.

https://github.com/celery/django-celery-results/blob/0a00044f25f7de7ff73c995727b69ff25c66f25d/t/unit/test_migrations.py#L26-L32

daadu: should we replace this with `call_command` thing i mentioned above? One advantage is that it is a higher level Django API - changes in Django internal will not affect the test - until only the command behaviour is changed.

auvipy: which part do you  want to replace with what? can you elaborate please?

auvipy: which part do you  want to replace with what? can you elaborate please?

daadu: Running `makemigrations django_celery_results --check --dry-run` command inside the test as mentioned above.

auvipy: ok

daadu: made those changes. Please review.

daadu: commit - https://github.com/celery/django-celery-results/pull/235/commits/113ac2e12e7eace5619f8855e0ff69a76b8f7996

auvipy:
can you check why CI is failing please

daadu:
@auvipy Looks like the CI is stuck.

ForthTurn:
I have a django model with a ForeignKey to TaskResult, and we set the `on_delete` to models.CASCADE. Maybe there is a problem after this change because the `_raw_delete` will not cause django's CASCADE delete?

auvipy:
can you please share more detail?

ForthTurn:
I have a django model named `MyTaskResult`, here is the code: 
``` python
class MyTaskResult(models.Model):
    result = models.OneToOneField(
        TaskResult,
        on_delete=models.CASCADE,
        verbose_name='TaskResult',
    )
    # other fields
```
and it works well when django-celery-results at 2.2.0.
But if update to 2.3.0 which the version include this mr, when I triggered TaskResult.objects.delete_expired():
``` python
from datetime import timedelta
from django_celery_results.models import TaskResult

t = timedelta(minutes=1)
TaskResult.objects.delete_expired(t)
```
I got this error: 
``` bash
django.db.utils.IntegrityError: (1451, 'Cannot delete or update a parent row: a foreign key constraint fails (..., CONSTRAINT `..._result_id_bbca4d3d_fk_django_ce` FOREIGN KEY (`result_id`) REFERENCES `django_celery_results_taskresult` (`id`))')"
```
I think the builtin task `celery.backend_cleanup` will also trigger this method and raise this exception.

In 2.2.0, `delete_expired()`calls `self.get_all_expired(expires).delete()`, and it will make django to delete the `MyTaskResult` before delete `TaskResult`.

ForthTurn:
> can you help to partially revert this manually? or provide a bettter mechanism??

Maybe can add a config key to switch? I saw it in [here](https://github.com/celery/django-celery-results/issues/232#issuecomment-912944517), but not in this mr

ForthTurn:
And I think the default behavior should be to use queryset's delete instead of `raw_delete`. `Queryset.delete` can be guaranteed to be correct. If the users can **confirm** they don't have a cascade ForeignKey or listen to the delete signal, they can switch the config to reduce the usage of memory.

auvipy:
> And I think the default behavior should be to use queryset's delete instead of `raw_delete`. `Queryset.delete` can be guaranteed to be correct. If the users can **confirm** they don't have a cascade ForeignKey or listen to the delete signal, they can switch the config to reduce the usage of memory.

I will be happy to accept a PR with partial revert of the PR[raw-delete part]

ForthTurn:
> I will be happy to accept a PR with partial revert of the PR[raw-delete part]

I will submit this at a later date :)

Issue #253, opened 2021-11-09
Store PeriodicTask's name


lvelvee:
Would there be a way to store the Periodic tasks name with the result ? @stuartaccent 's method only store the task name like demoapp.tasks.add, cannot store table 'django_celery_beat_periodictask''s name field.

_Originally posted by @guohongze in https://github.com/celery/django-celery-results/issues/7#issuecomment-300369705_

lvelvee:
I'm glad to make some contributions, but the two repositories ( `django-celery-results` and `django-celery-beat` ) have to make some changes simultaneously, is this possible to be merged?

auvipy:
what changes are you willing to make? start with one repo first

lvelvee:
This is part of the patch of my previous project:

```diff

--- a/vendors/django-celery-beat/django_celery_beat/schedulers.py	(revision 7c7af2034bb51d1143544a5f47604cb5c74b2633)
+++ b/vendors/django-celery-beat/django_celery_beat/schedulers.py	(revision 61ac7c03c7ecc6e7e02cb84c1b423f5be1cb84be)
@@ -85,6 +85,7 @@
             self.options['expires'] = getattr(model, 'expires_')
 
         self.options['headers'] = loads(model.headers or '{}')
+        self.options['periodic_task_name'] = model.name
 
         self.total_run_count = model.total_run_count
         self.model = model


--- a/vendors/django-celery-results/django_celery_results/backends/database.py	(revision 7c7af2034bb51d1143544a5f47604cb5c74b2633)
+++ b/vendors/django-celery-results/django_celery_results/backends/database.py	(revision 61ac7c03c7ecc6e7e02cb84c1b423f5be1cb84be)
@@ -39,6 +39,7 @@
         )
 
         task_name = getattr(request, 'task', None)
+        periodic_task_name = getattr(request, 'properties', None)['periodic_task_name']
         worker = getattr(request, 'hostname', None)
 
         # Get input arguments
@@ -72,6 +73,7 @@
             traceback=traceback,
             meta=meta,
             task_name=task_name,
+            periodic_task_name=periodic_task_name,
             task_args=task_args,
             task_kwargs=task_kwargs,
             worker=worker,


--- a/vendors/django-celery-results/django_celery_results/models.py	(revision 7c7af2034bb51d1143544a5f47604cb5c74b2633)
[...]

PR #261, opened 2021-11-18
feat: add periodic_task_name

lvelvee:
#253


lvelvee:
related: https://github.com/lvelvee/django-celery-beat/tree/feat-periodic-task-name

auvipy (changes requested):
please check the test failures

auvipy:
some documentation change is require i guess

auvipy:
I am curious what extra benefits we get by adding this? what the existing 2 packages are missing?

lvelvee:
> I am curious what extra benefits we get by adding this? what the existing 2 packages are missing?

If a `CeleryTask` may correspond to more than one `PeriodicTask`, it will not be possible to distinguish the source of the `PeriodicTask` because django-celery-results lacks the name field of the `PeriodicTask`.

For example, if I have two `PeriodicTask` named `periodic_a` and `periodic_b`, both of which call the `test` task of celery at regular intervals, without this patch, the results will only be known as the results of `test`, and there will be no way to distinguish whether the task originated from `periodic_a` or `periodic_b`.

lvelvee:
@auvipy Is it still possible to be merged?

auvipy (approved):
i guess some change in celery beat is also required?

lvelvee:
I'll make the pr later

PR #263, opened 2021-11-20
Add Simplified Chinese translation strings and fix comment typo.

ifmos:
As a Chinese user, I noticed that there is no Simplified Chinese translation, so I added one by myself. And then fix a comment typo btw. Hope to be merged.

### DIFF
--- a/locale/zh_Hans/LC_MESSAGES/django.po
+++ b/locale/zh_Hans/LC_MESSAGES/django.po
@@ -1,0 +1,146 @@
+# Simplified Chinese translation strings for django-celery-results.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as django-celery-results.
+# <acwzy@live.com>, 2021.
+#
+#, fuzzy
+msgid ""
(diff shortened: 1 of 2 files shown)


locale/zh_Hans/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,146 @@
+# Simplified Chinese translation strings for django-celery-results.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as django-celery-results.
+# <acwzy@live.com>, 2021.
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version:\n"
```
auvipy, line 9: need another chinese speaker to verify

auvipy (changes requested):
thanks for the PR

locale/zh_Hans/LC_MESSAGES/django.po
lvelvee: the translation is ok

auvipy (approved):

Issue #264, opened 2021-11-20
No locale files found installed by pip.


ifmos:
It's a nice project and help me a lot. I tried to make my django website localized and run `compilemessages` command. But I can not find locale files when install it with pip. How can I do with this problem?
btw, I created a pull request to add Simplified Chinese translations #263 , hope that is useful for i18n.

PR #265, opened 2021-11-21
Fix locale in dir tree.

ifmos:
Hi guys,
I have reported an issue before in #264 , which means there is no locale files after build. And I found out that it works fine in [`django-celery-beat`](https://github.com/celery/django-celery-beat), so I checked out this commit celery/django-celery-beat#395 on it and make it work on this repo. 
I have tested it on my local machine using build command `python setup.py sdist bdist_wheel`, and it worked out with locale files put in lib dir. Wonder if it works on release.


auvipy (approved):

auvipy:
what about this aswell https://github.com/celery/django-celery-beat/commit/079b2f059e81fd33dac8468c6eb5c299d47b533a

Issue #285, opened 2022-02-07
MySQL error on migrate persists


benselme:
I'm aware of `DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH`.

However migration `0004_auto_20190516_0412.py` still fails with `django.db.utils.OperationalError: (1709, 'Index column size too large. The maximum column size is 767 bytes.')` because it's trying to create the `task_name` column which is 255-characters long and indexed.

auvipy:
can you elaborate more on the versions of everything you are using?

benselme:
The database is MariaDB 10.1.48, but I deploy on an older version, 5.5.60

benselme:
That would be MySQL 5.6 which is equivalent to MariaDB 10.1. For MariaDB >= 10.2.2, the server setting `innodb_large_prefix` defaults to `ON`, which should remove the limitation that causes the crash.

auvipy:
> but I deploy on an older version, 5.5.60

sorry can't accept any contribution for unsuported versions :)

benselme:
You misread what I wrote. Django 2.2 supports MySQL 5.6 and so also MariaDB 10.1, on which the problem happens.

PR #287, opened 2022-02-08
Fix field size for MySQL (#285)

benselme:
MySQL/MariaDB do not accept indexed fields larger than 191 chars when using `utf8mb4` charset. This makes migrations fail because the `task_name` field is 255-chars long and indexed.

This issue was fixed for the `task_id` field using a new `DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH` Django setting. This PR applies the same fix for field `task_name`.


auvipy (approved):

benselme:
@auvipy Thank you!

Issue #119, opened 2019-11-28
Feature: retry storage attempt on db exceptions


asfaltboy:
Similarly to how redis backend retries a connection in case server is unreachable, the idea is to implement a similar retry make storing a result more resilient to failures.

<details>
  <summary>An example implementation</summary>
  
We came up with the following workaround class while this feature is being worked on:
  
  ```python
from functools import partial

from celery.utils.log import get_logger
from celery.utils.time import humanize_seconds
from django_celery_results.backends import DatabaseBackend
from kombu.utils.functional import retry_over_time
from psycopg2 import DatabaseError as Psycopg2DatabaseError

from django.db.utils import DatabaseError

E_LOST = 'Connection to Django DB lost: Retry (%s/%s) %s.'
EXCEPTIONS_TO_CATCH = (DatabaseError, Psycopg2DatabaseError)
MAX_RETRIES = 20

logger = get_logger(__name__)


class RetryingDatabaseBackend(DatabaseBackend):
    """
    A sub-class of django_celery_results.backends.DatabaseBackend, this backend
    provides a retry functionality to the storage facility, guarding against
    a short-lived failure (like a "Temporary failure in name resolution").

    An additional benefit of this, is to reduce the rate of fetching the tasks
    from the broker (which are likely to fail at this stage).

    It's more or less copied as-is from the same feature in the redis backend.
    """

    retry_policy = {
        'max_retries': MAX_RETRIES,
        'interval_start': 0,
        'interval_step': 1,
        'interval_max': 1,
    }

    def ensure(self, fun, args, kwargs):
        retry_policy = dict(self.retry_policy)
        max_retries = retry_policy.get('max_retries')
        return retry_over_time(
            fun=fun,
            catch=EXCEPTIONS_TO_CATCH,
            args=args,
            kwargs=kwargs,
            errback=partial(self.on_connection_error, max_retries),
            **retry_policy,
        )

    def on_connection_error(self, max_retries, exc, intervals, retries):
        tts = next(intervals)
        logger.error(
            E_LOST, retries, max_retries or 'Inf', humanize_seconds(tts, 'in ')
        )
        return tts

    def _store_result(self, *args, **kwargs):
        return self.ensure(super()._store_result, args, kwargs)
  ```
  
</details>

If the above feature is useful, we can submit a pull-request to add this to the `DatabaseBackend` class.

PR #290, opened 2022-02-24
Add retry logic for Django and Psycopg2 InterfaceErrors

juspence:
This PR implements https://github.com/celery/django-celery-results/issues/119 in a slightly different way. It adds retry logic only for django.db.utils.InterfaceError and psycopg2.InterfaceError, which can be errors like "connection already closed".

It does not retry any DatabaseError, which can be errors like "numeric value out of range" or "foreign key check failed". See https://www.psycopg.org/docs/module.html#exceptions for more info.

exception_safe_to_retry() is overridden from the parent class: https://docs.celeryproject.org/en/stable/_modules/celery/backends/base.html#BaseBackend

This method is called when an exception is raised in _store_result() and always_retry / result_backend_always_retry is set to True. If the exception is an InterfaceError, the connection is closed as a side effect of calling exception_safe_to_retry().

Then the retry logic implemented in the parent class continues, and Django will automatically open a new connection for the next attempt to store the task result or get the task metadata.

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -7,12 +7,16 @@ from celery.exceptions import ChordError
 from celery.result import GroupResult, allow_join_result, result_from_tuple
 from celery.utils.log import get_logger
 from celery.utils.serialization import b64decode, b64encode
-from django.db import transaction
+from django.db import connection, transaction
+from django.db.utils import InterfaceError
 from kombu.exceptions import DecodeError
+from psycopg2 import InterfaceError as Psycopg2InterfaceError
 
 from ..models import ChordCounter
 from ..models import GroupResult as GroupResultModel
 from ..models import TaskResult
+
+EXCEPTIONS_TO_CATCH = (InterfaceError, Psycopg2InterfaceError)
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
I would like to see unit tests for this change

auvipy:
I want to add this to next release, can you please address the review comment?

juspence:
@auvipy Yes, I am working on this today since I had other tasks to finish Friday.

auvipy (approved):

django_celery_results/backends/database.py
```diff
@@ -7,13 +7,17 @@
 from celery.result import GroupResult, allow_join_result, result_from_tuple
 from celery.utils.log import get_logger
 from celery.utils.serialization import b64decode, b64encode
-from django.db import transaction
+from django.db import connection, transaction
+from django.db.utils import InterfaceError
 from kombu.exceptions import DecodeError
+from psycopg2 import InterfaceError as Psycopg2InterfaceError
```
intgr, line 13: This adds a hard dependency on psycopg2, but django-celery-results can be used with other databases as well.

I have opened PR #292 to fix this.

PR #292, opened 2022-03-01
Remove hard dependency on psycopg2

intgr:
One would reasonably expect django-celery-results to work with other databases without installing psycopg2.

The hard dependency was added in PR #290, probably unintentional without fully understanding this aspect.

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -10,13 +10,18 @@ from celery.utils.serialization import b64decode, b64encode
 from django.db import connection, transaction
 from django.db.utils import InterfaceError
 from kombu.exceptions import DecodeError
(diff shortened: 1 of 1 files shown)


intgr:
As a work-around, I have added `psycopg2-binary` dependency to my project, so it's not a big deal.

auvipy:
> As a work-around, I have added `psycopg2-binary` dependency to my project, so it's not a big deal.

can we add that to this package as well in this PR?

auvipy:
what confuses me is the test failures with https://github.com/celery/django-celery-results/pull/294/files

auvipy (approved):

saemideluxe:
Thanks for fixing this. I just encountered the issue after an upgrade. Any chance that there will soon be a new pip release including the fix?

Issue #323, opened 2022-06-21
Make Celery Result Backend work when there are multiple database but no default


nofalx:
I detect this issue occurring after i set up multiple database with no default DB in 

```
ImproperlyConfigured 
celery.backend_cleanup
settings.DATABASES is improperly configured. Please supply the ENGINE value. Check..
```

My Django setting as below

```
CELERY_QUEUE = "queue_one"
CELERY_RESULT_BACKEND = "django-db"
DATABASES = {
    "default": {},
    "main": {
        "ENGINE": "django.db.backends.postgresql",
        "NAME": "..",
        "USER": "..",
        "PASSWORD": "..",
        "HOST": "..",
        "PORT": "..",  
    },
    "datalake": {
        "ENGINE": "timescale.db.backends.postgresql",
        "NAME": "..",
        "USER": "..",
        "PASSWORD": "..",
        "HOST": "..",
        "PORT": "..",  
    },
}
DATABASE_ROUTERS = [
    "app.db_routers.TimeScaleRouter",
    "app.db_routers.MainRouter",
]
``` 
  
The issue occurs because we are trying to perform an atomic transaction at
https://github.com/celery/django-celery-results/blob/16a18d9af18635b025523cec0d0c32788920011e/django_celery_results/managers.py#L90
  
The line `transaction.atomic()` will always point to the default db if you look at the django source code at:
https://github.com/django/django/blob/2887b9f67cadc5295ef6a0574de2c2c8fdd66905/django/db/transaction.py#L315

I tried to read the [documentation](https://docs.celeryq.dev/en/latest/userguide/configuration.html#conf-result-backend) and see if we can specify which DATABASES we can manually specify which db but i failed to find it, in case of settings above `main` or `datalake`.

Im assuming the database is determined using the `DATABASE_ROUTERS`, in this case can we somehow inject the db name into the atomic transaction like this:

`transaction.atomic(using=self.db)`

auvipy:
I think https://github.com/celery/django-celery-beat/pull/451/files is somehow related to what you are thinking

auvipy:
released

PR #324, opened 2022-06-21
Fix atomic transaction not routing to the the correct DB

nofalx:
This to fix the issue explained at: https://github.com/celery/django-celery-results/issues/323

### DIFF
--- a/django_celery_results/managers.py
+++ b/django_celery_results/managers.py
@@ -87,7 +87,7 @@ class ResultManager(models.Manager):
 
     def delete_expired(self, expires):
         """Delete all expired results."""
-        with transaction.atomic():
+        with transaction.atomic(using=self.db):
             raw_delete(queryset=self.get_all_expired(expires))
 
 


auvipy (changes requested):
can you also add some tests for the change?

nofalx:
I have added a new test class and a db router as these are required to test this specific scenario

using `make test`

The test passes after the edits to the `manager.py` file

without any changes to `manager.py` the test fails with the below error:
```
FAILED t/unit/test_models.py::test_ModelsOnSecondaryDbOnly::test_operations_with_atomic_transactions - AssertionError: Database connections to 'default' are not allowed in this test. Add 'default' to t.unit.test_models.test_ModelsOnSecondaryDbOnly.databases to ensure proper test isolation and silence this failure.
```
Screenshot from 2022-06-21 13-50-10

t/proj/db_routers.py
```diff
@@ -0,0 +1,26 @@
+class AlwaysSecondaryDbRouter:
```
auvipy, line 1: why always Secondary DB router?

t/unit/test_models.py
```diff
@@ -209,3 +209,21 @@ class TransactionError(Exception):
                 raise TransactionError()
         except TransactionError:
             pass
+
+
+@pytest.mark.usefixtures('depends_on_current_app')
+class test_ModelsOnSecondaryDbOnly(TransactionTestCase):
+    """
+    These tests will fail with the below error incase we
```
auvipy, line 217: can you explain this part a bit more please?

t/proj/db_routers.py
nofalx: I removed this part and simplified the test

t/unit/test_models.py
nofalx: We will try to run the method `delete_expired` from the Custom Manager `ResultManager` where the bug occurs.

The issue is that if the manager was directed to use a db other than the "default" db, the method `delete_expired` would run the operation on the "default" anyways. Since the function does not return anything and is a delete operation I can not use assert

My method of testing is the below: (let me know if there is a better way)

I specify the allowed db to test on as "secondary" only. If the test tries to use the "default" db it will throw error:
https://github.com/nofalx/django-celery-results/blob/fix-323/t/unit/test_models.py#L231:L232

I specify the `TaskResult` and the `GroupResult` manager which extend `ResultManager` to use "secondary" db using `.db_manager("secondary")`. 

https://github.com/nofalx/django-celery-results/blob/fix-323/t/unit/test_models.py#L235:L240

Before the fixes this test would fail as `transaction.atomic()` would try to use the "default" db. After the fix the method doesnt use "default" db anymore and the test pass.

auvipy (approved):

Issue #326, opened 2022-06-30
Task name and positional arguments don't appear in the results


ddahan:
No matter how I run a task, I can't get the task name nor the parameters display in the results:

<img width="724" alt="image" src="https://user-images.githubusercontent.com/1933516/176698922-de7220e2-6bfb-4561-a48b-0f3259872de8.png">

Here is my task (I tried with `bind=True` too):

```python
@app.task(name="core.add")
def add(x, y):
    return x + y
```

To run the task, I tried:
- In a shell: `add.delay(2, 10)`
- In a shell: `add.apply_async(args=[2, 10])`
- Using django-celery-beat, setting a new `PeriodicTask` object with position arguments set to `[2, 10]`

Same issue every time. What did I do wrong? Thanks.

cuiliang0302:
![image](https://user-images.githubusercontent.com/42726109/176804027-89f15770-494d-4a03-a0a7-c5dff96bccc8.png)

me too

cuiliang0302:
![image](https://user-images.githubusercontent.com/42726109/176806339-373faa65-714d-4dfb-9b72-ae618c24325e.png)

There is no problem with the replacement version 2.3.1

auvipy:
v2.4.0 released and v2.4.1 is coming

auvipy:
@[AmitPhulera](https://github.com/celery/django-celery-results/commits?author=AmitPhulera) can you please verify the regression?

valberg:
We are experiencing the same issues since 2.4.0

auvipy:
is it https://github.com/celery/django-celery-results/commit/ad508fe3433499e5fc94645412d911e174863f28 ?

andreasgou:
After 2.4.0 upgrade (from 2.2.0), I see no tasks and no Periodic task name no positional arguments.
The upgrade was made in an attempt to solve a serious blocker with Oracle DB (DPI-1010 no connection) happening at random times. (I hope it is solved)

By when do you think we'll get the results?

auvipy:
you can try 2.3.1

valberg:
@auvipy It works as intended in e174c9978c66f7d9c753d9332e2ec66c8abe1c1a and in fact ad508fe3433499e5fc94645412d911e174863f28 seems to be the culprit.

auvipy:
[...]

PR #328, opened 2022-07-07
Create getting_started.rst

valberg:
Trying to clear up the confusion from #326 

Rendered it looks like this:

image

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -1,0 +1,64 @@
+Getting started
+===============
+
+To use :pypi:`django-celery-results` with your project you need to follow these steps:
+
+#. Install the :pypi:`django-celery-results` library:
+
+    .. code-block:: console
+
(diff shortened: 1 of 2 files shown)


AmitPhulera:
This looks good @valberg 🎉 
Thanks for adding these.

tisdall:
I know this is copied from the celery docs, but I'm wondering if `CELERY_CACHE_BACKEND = 'django-cache'` is actually correct...  It seems like you need to use `CELERY_RESULT_BACKEND = 'django-cache'` to tell it to use the cache backend defined in `django-celery-results` and then `CELERY_CACHE_BACKEND` to specify which cache to use.

Regardless of the above, it might be good to include the `CELERY_RESULT_EXTENDED` documentation here too:  https://docs.celeryq.dev/en/stable/django/first-steps-with-django.html

PR #351, opened 2022-12-18
Add Django to install_requires and test on Django 4.1/Python 3.11

gopackgo90:
Also removed some remaining django 2.2 mentions


.github/workflows/test.yaml
```diff
@@ -13,8 +13,17 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                python-version: [ 3.8, 3.9, "3.10" ]
-                django: [ 32, 40 ]
+                python-version: [ 3.7, 3.8, 3.9, "3.10", "3.11" ]
```
auvipy, line 16: i would like to add pypy-3.9 instead of python 3.7

auvipy (changes requested):

.github/workflows/test.yaml
gopackgo90: I added pypy-3.9, but I think it's a good idea to keep python 3.7 in here as it's the minimum supported Python version supported by this library currently and most likely to break if someone accidentally adds a language feature that's python 3.8+.

auvipy: yeah ok

auvipy (approved):

PR #358, opened 2023-01-16
Ensure `delete_expired` uses writable database.

### LINKED ISSUE #357: The `delete_expired` method of ResultManager sometimes selects read-only database
In a django project with a database router and read-only databases, the [`delete_expired`](https://github.com/celery/django-celery-results/blob/main/django_celery_results/managers.py#L87) method may try to perform `raw_delete()` using a read-only database, resulting in an Internal Error: "ReadOnlySqlTransaction - cannot execute DELETE in a read-only transaction".

[Relevant code](https://github.com/celery/django-celery-results/blob/main/django_celery_results/managers.py#L87):

```python
def delete_expired(self, expires):
    """Delete all expired results."""
    with transaction.atomic(using=self.db):
        raw_delete(queryset=self.get_all_expired(expires))
```

My Database Router:

```python
import random

from django.conf import settings


class CustomRouter:
    def db_for_read(self, model, **hints):
        """
        Return either default or one of the replicas
        """

        db_read = [
            "default",
        ]

        if settings.REPLICA1_URL is not None:
            db_read.append("replica1")
        if settings.REPLICA2_URL is not None:
            db_read.append("replica2")

        return random.choice(db_read)

    def db_for_write(self, model, **hints):
        # Always return the default database
        return "default"

    def allow_relation(self, obj1, obj2, **hints):
        return True

    def allow_migrate(self, db, app_label, model_name=None, **hints):
        return True
```

Occasionally, django-celery-results will attempt to conduct `delete_expired` using one of my replica databases:

`connection: <DatabaseWrapper vendor='postgresql' alias='replica1'>`

Recommend changing the `using` argument of the atomic transaction in `delete_expired` to explicitly use the writable database specified in the router.

jacklinke:
Update `delete_expired` method of ResultManager to use writable database.

Resolves #357

### DIFF
--- a/django_celery_results/managers.py
+++ b/django_celery_results/managers.py
@@ -68,8 +68,11 @@ class ResultManager(models.Manager):
                 if isolation == 'REPEATABLE-READ':
                     warnings.warn(TxIsolationWarning(W_ISOLATION_REP.strip()))
 
+    def database_for_write(self):
+        return router.db_for_write(self.model)
+
     def connection_for_write(self):
-        return connections[router.db_for_write(self.model)]
+        return connections[self.database_for_write()]
 
     def connection_for_read(self):
         return connections[self.db]
@@ -86,7 +89,7 @@ class ResultManager(models.Manager):
 
     def delete_expired(self, expires):
         """Delete all expired results."""
-        with transaction.atomic(using=self.db):
+        with transaction.atomic(using=self.database_for_write()):
             raw_delete(queryset=self.get_all_expired(expires))
 
 


auvipy (changes requested):
you will need to update the tests as well

auvipy (changes requested):
please add / update tests to make the CI green.

django_celery_results/managers.py
```diff
@@ -86,7 +89,7 @@ def get_all_expired(self, expires):
 
     def delete_expired(self, expires):
         """Delete all expired results."""
-        with transaction.atomic(using=self.db):
+        with transaction.atomic(using=self.database_for_write()):
```
auvipy, line 92: also resolve the merge conflict please?

auvipy (changes requested):

matmunn:
@jacklinke any joy on this one? I'm running into the same issue that you are

PR #366, opened 2023-02-24
Meta injection

rodrigondec:
# Resume
- related to #37 

The issue #37 points to save data using the arg `meta` from the method `update_state`, but the `meta` attribute is renamed to `result` when it is passed to `self.backend.store_result`.

As explained here in this comment https://github.com/celery/django-celery-results/issues/37#issuecomment-385203222.

> https://github.com/celery/celery/blob/cb4c3256326e33f005c63d48520d0abb5e898151/celery/app/task.py#L982-L994
```python
    def update_state(self, task_id=None, state=None, meta=None, **kwargs):
        """Update task state.

        Arguments:
            task_id (str): Id of the task to update.
                Defaults to the id of the current task.
            state (str): New state.
            meta (Dict): State meta-data.
        """
        if task_id is None:
            task_id = self.request.id
        self.backend.store_result(
            task_id, meta, state, request=self.request, **kwargs)
```

> https://github.com/celery/celery/blob/cb4c3256326e33f005c63d48520d0abb5e898151/celery/backends/base.py#L513-L522
```python
    def store_result(self, task_id, result, state,
                     traceback=None, request=None, **kwargs):
        """Update task state and result.

        if always_retry_backend_operation is activated, in the event of a recoverable exception,
        then retry operation with an exponential backoff until a limit has been reached.
        """
        result = self.encode_result(result, state)


        retries = 0
```

# Problem
So the argued `meta` arg on `store_result` is not really arbitrary metadata. It's actually the result of the Task.

# Solution
Use the TaskResult.meta to store arbitrary data.

# How to inject?
I implemented the meta injection through the Task Celery Request. This way I don't mess around with too many implementations and function signatures.

PR #407, opened 2023-09-03
fix: allow for custom states in status field

ericswpark:
From https://github.com/celery/django-celery-results/pull/366#issuecomment-1586237389

Thanks to @cdevacc1 for the idea. I just decided to write up the PR as it only takes a couple of minutes. This should resolve the bug where custom states are not shown at all in the admin panel.


django_celery_results/models.py
```diff
@@ -51,7 +51,6 @@ class TaskResult(models.Model):
                     'used with the task'))
     status = models.CharField(
         max_length=50, default=states.PENDING,
-        choices=TASK_STATE_CHOICES,
```
auvipy, line 54: I think it is a breaking change. ma I know what benefits removing this offer and what else we can do without removing this? and thanks for handling the issue

auvipy (changes requested):

auvipy:
And tests are already failing you can see that

django_celery_results/models.py
ericswpark: Celery allows for settings statuses that are not part of `TASK_STATE_CHOICES`, such as `PROGRESS`. However, when set to a non-standard value, `django-celery-results` will fail to show the status, displaying it as `-` instead.

ericswpark:
@auvipy the tests should now pass with the added migration. Can I please request a review? Thanks!

django_celery_results/models.py
auvipy: I would love to see examples and screenshots if possible

auvipy (changes requested):

django_celery_results/models.py
auvipy: as far I can see the states are coming from celery https://github.com/celery/django-celery-results/blob/main/django_celery_results/models.py#L5

ericswpark: https://github.com/celery/celery/blob/3b20010020cf810a0ba40a27d6d4e83210d4a5e2/celery/states.py#L103-L104

In the celery/states file, you can see that they allow for "custom states" that are not defined in the file.

Even the documentation suggests this with the `PROGRESS` state: https://docs.celeryq.dev/en/stable/userguide/calling.html#on-message

If a developer followed the above userguide provided by Celery, then DCR will show a `-` instead of `PROGRESS` because of this.

auvipy, line 54: the problem with this change I can see is, people using this choice field in existing projects will face breaking change. so we have to find a solution which is backward compatible and existing apps won't break.

auvipy (changes requested):

django_celery_results/models.py
ericswpark: @auvipy wouldn't it still let them use the choices? From my understanding removing the `choices` parameter just unrestricts the choices from `TASK_STATE_CHOICES`. What kind of breaking change are you talking about here?

ericswpark:
@auvipy I'm not sure what breaking changes you're concerned about -- removing the line should allow for custom states while letting those who originally used the `TASK_STATE_CHOICES` for choices keep their data with no changes necessary. Please let me know if there's anything else you need or any other changes that should be made before the PR is merged or rejected. Thanks.

django_celery_results/migrations/0012_alter_taskresult_status.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 4.2.5 on 2023-09-04 14:56
```
auvipy, line 1: I think we need to remove this and regenerate again after pulling from main branch

auvipy (changes requested):

django_celery_results/migrations/0012_alter_taskresult_status.py
ericswpark: Done; I've rebased and created new migrations.

ericswpark:
@auvipy may this please be reviewed before it becomes stale? I'd like to avoid having to rebase it over and over and this PR's been sitting in my list since last year. Thanks!

auvipy:
> @auvipy I'm not sure what breaking changes you're concerned about -- removing the line should allow for custom states while letting those who originally used the `TASK_STATE_CHOICES` for choices keep their data with no changes necessary. Please let me know if there's anything else you need or any other changes that should be made before the PR is merged or rejected. Thanks.

my only concern is if existing users are going to miss anything by the removal of the `TASK_STATE_CHOICES`

ericswpark:
> my only concern is if existing users are going to miss anything by the removal of the `TASK_STATE_CHOICES`

Please see previous comments. It should not affect existing users as it only _limits_ what choices can go inside the field. Removing it expands it to allow for other values, not just the ones in `TASK_STATE_CHOICES`.

auvipy:
@AllexVeldman can you review this as well, please?

AllexVeldman (approved):
LGTM

auvipy (changes requested):
can you please make the integration tests pass for this PR? with the change, may be it would be good idea to add/adjust additional tests

AllexVeldman:
@ericswpark I think this will cover it:

```python
def test_custom_state(self):
        tid = uuid()

        assert self.b.get_status(tid) == states.PENDING

        self.b.store_result(tid, state="Progress", result={"progress": 10})
        assert self.b.get_status(tid) == "Progress"
        assert self.b.get_result(tid) == {"progress": 10}

        self.b.mark_as_done(tid, 42)
        assert self.b.get_status(tid) == states.SUCCESS
        assert self.b.get_result(tid) == 42
```

Note that this does not fail on `main` since the backend never calls `TaskResult.full_clean()` so model validation seems to never be executed.

@auvipy This also means this change only changes the field in the Django admin pages from a dropdown to a text field, no functional changes here.

ericswpark:
@AllexVeldman thank you for the tests, I tried adding them to the tests directory under the tests for the models but am not sure that they belong there. I was hoping GitHub Actions/CI would kick in and verify but it seems like the workflow needs to be approved by a maintainer for them to run.

ericswpark:
Ach, I think the migrations are failing because this PR is so old. At some point another migration file was pushed to main and then this PR was rebased on top without recreating migrations. Will recreate migrations and push again shortly.

django_celery_results/migrations/0013_alter_taskresult_status.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 5.0.6 on 2024-06-19 09:26
```
auvipy, line 1: you can delete this file and regenerate a new one with django 5.1.x

auvipy (changes requested):

AllexVeldman:
> I tried adding them to the tests directory under the tests for the models but am not sure that they belong there.

@ericswpark I should have mentioned :) this would go into `/t/unit/backends/test_database.py` since it tests setting a custom state, not specifically how the state is stored in the db model.

django_celery_results/migrations/0013_alter_taskresult_status.py
ericswpark: Done!

ericswpark:
Sorry about that, I forgot to sync my `main` branch on my fork. The new migration should be based off of the last one (`0014`) created on `main`.

auvipy (approved):

PR #413, opened 2023-11-22
fix: header_result.id duplication exception

hann-wang:
Chaining multiple groups together with a single task will be upgraded to a chord, which further leads to a unique constraint violation on the `django_celery_results_chordcounter.group_id` key. Perhaps we should change the `create` method to `update_or_create` method?


auvipy:
why it is closed?

hann-wang:
I am investigating why the same `group_id` is called multiple times by celery canvas. A temporary solution might be adding up all `count`s instead of resetting to new one.

PR #425, opened 2024-03-05
add django 5.0 to CI/test matrix and project classifiers

charlesobrien:
Adding Django 5.0 to CI/test matrix and bumping some other Django minor versions

PR #426, opened 2024-03-20
Change default_auto_field to BigAutoField

### LINKED ISSUE #307: Reaching maximum value for `TaskResult` primary key field causes task failures
When using `django-celery-results` with the Postgres backend, it's possible to reach the maximum integer value for the primary key column, and this causes every task to fail. 

We're using Python 3.9.7, Django 2.2.28, Django Celery Results 2.2.0, Postgres 12.

Traceback:
```
TaskResult.DoesNotExist: TaskResult matching query does not exist.
  File "django/db/models/query.py", line 538, in get_or_create
    return self.get(**kwargs), False
  File "django/db/models/query.py", line 406, in get
    raise self.model.DoesNotExist(
SequenceGeneratorLimitExceeded: nextval: reached maximum value of sequence "django_celery_results_taskresult_id_seq" (2147483647)

  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
DataError: nextval: reached maximum value of sequence "django_celery_results_taskresult_id_seq" (2147483647)

  File "billiard/pool.py", line 1796, in safe_apply_callback
    fun(*args, **kwargs)
  File "celery/worker/request.py", line 571, in on_failure
    self.task.backend.mark_as_failure(
  File "celery/backends/base.py", line 171, in mark_as_failure
    self.store_result(task_id, exc, state,
  File "celery/backends/base.py", line 482, in store_result
    self._store_result(task_id, result, state, traceback,
  File "django_celery_results/backends/database.py", line 66, in _store_result
    self.TaskModel._default_manager.store_result(
  File "django_celery_results/managers.py", line 46, in _inner
    return fun(*args, **kwargs)
  File "django_celery_results/managers.py", line 168, in store_result
    obj, created = self.using(using).get_or_create(task_id=task_id,
  File "django/db/models/query.py", line 541, in get_or_create
    return self._create_object_from_params(kwargs, params)
  File "django/db/models/query.py", line 575, in _create_object_from_params
    obj = self.create(**params)
  File "django/db/models/query.py", line 422, in create
    obj.save(force_insert=True, using=self.db)
  File "django/db/models/base.py", line 743, in save
    self.save_base(using=using, force_insert=force_insert,
  File "django/db/models/base.py", line 780, in save_base
    updated = self._save_table(
  File "django/db/models/base.py", line 873, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "django/db/models/base.py", line 910, in _do_insert
    return manager._insert([self], fields=fields, return_id=update_pk,
  File "django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 1186, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "cachalot/monkey_patch.py", line 37, in inner
    return original(compiler, *args, **kwargs)
  File "cachalot/monkey_patch.py", line 113, in inner
    return original(write_compiler, *args, **kwargs)
  File "django/db/models/sql/compiler.py", line 1377, in execute_sql
    cursor.execute(sql, params)
  File "cachalot/monkey_patch.py", line 137, in inner
    return original(cursor, sql, *args, **kwargs)
  File "django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "django/db/backends/utils.py", line 76, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "django/db/utils.py", line 89, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
```

The fix was to manually `ALTER SEQUENCE` and `ALTER COLUMN` to `bigint`. After this, tasks continued to work as expected.

Although our problem is now solved, I'm not sure if there was a way to prevent this, or at least handle the issue in a better way.

browniebroke:
Fix #307 

Or at least make the issue much less frequent...


browniebroke:
I don't understand the test failure, they seem unrelated with my changes... Are they broken due to an external factor or am I missing something obvious?

charlesobrien:
@browniebroke They're failing because Pytest was unconstrained and breaking changes in Pytest 8 (released in Jan) are blowing up the tests.

My PR https://github.com/celery/django-celery-results/pull/425 to add Django 5 to CI also places a constraint on pytest. Hoping it might be able to get reviewed/merged soon.

browniebroke:
Thanks a lot! Will wait for a bit then.

django_celery_results/apps.py
```diff
@@ -12,4 +12,4 @@ class CeleryResultConfig(AppConfig):
     name = 'django_celery_results'
     label = 'django_celery_results'
     verbose_name = _('Celery Results')
-    default_auto_field = 'django.db.models.AutoField'
+    default_auto_field = 'django.db.models.BigAutoField'
```
auvipy, line 15: I'm curious, if this change is backward compatible?

auvipy (changes requested):

django_celery_results/apps.py
browniebroke: I _think_ it's backwards compatible. It's changing the DB type from an integer field to a 64-bit integer, increasing its range. 

However, I have a few concerns for users of the library:

- I'm not sure if there are any performances considerations when applying this migration on a table with a large number of rows. Would the `migrate` script takes a while to run? What kind of locks does it need? How does it varies between databases? 
- The django docs have [a large note](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-DEFAULT_AUTO_FIELD) about the change, and if folks have a many to many field that link to one of the tables, they'll need to do an extra migration:
   > Unfortunately, the primary keys of existing auto-created through tables cannot currently be updated by the migrations framework.

So it might be worthy of a major bump to signal these pitfalls...

browniebroke:
Applied a similar migration today in a project and learned the hard way that it's not as simple as that 😔

auvipy:
yeah you are right

Issue #422, opened 2024-03-01
Getting exception "select_for_update cannot be used outside of a transaction"


rehbarkhan:
django-celery-results is not able to save the tasks results, getting below error. :

```
{"exc_type": "TransactionManagementError", "exc_message": ["select_for_update cannot be used outside of a transaction."], "exc_module": "django.db.transaction"}

```

My App Config:
- django==4.2.*
- django-celery-beat==2.5.0
- django-celery-results==2.5.1

My database config: 
```
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.mysql',
        'NAME': 'production',
        'USER': 'production_user',
        'PASSWORD': env('MYSQL_PASSWORD'),
        'HOST': env('MYSQL_HOST'),
        'CONN_MAX_AGE': 600,
        # 'ATOMIC_REQUESTS': True,
    }
}

DATABASES['production_db'] = DATABASES['default']
```

I have tried to enable "ATOMIC_REQUESTS: True" but no luck, still getting the TransactionManagementError

auvipy:
would you mind checking this please https://stackoverflow.com/questions/25451087/django-select-for-update-cannot-be-used-outside-of-a-transaction ?

rehbarkhan:
I have tried this, but it didn't work.
I went to the django-celery-results code base and updated the blow code 
```
def on_chord_part_return(self, request, state, result, **kwargs):
        """Called on finishing each part of a Chord header"""
        tid, gid = request.id, request.group
        if not gid or not tid:
            return
        call_callback = False
        with transaction.atomic():
            # We need to know if `count` hits 0.
            # wrap the update in a transaction
 ```
with 
```
def on_chord_part_return(self, request, state, result, **kwargs):
        """Called on finishing each part of a Chord header"""
        tid, gid = request.id, request.group
        if not gid or not tid:
            return
        call_callback = False
        with transaction.atomic(using = 'production_db'):
            # We need to know if `count` hits 0.
            # wrap the update in a transaction
```

I have passed ``` using='production_db' ``` to atomic method.

After making these changes, results are getting stored in the django database however after 30 to 40mins, celery is existing automatically.

PR #427, opened 2024-04-05
Fix atomic transaction not routing to the the correct DB in DatabaseBackend.on_chord_part_return transaction.atomic

gianbot:
Hi,
 when using chord with Database result backend and multiple databases with a Database Router we encounter this error:
`select_for_update cannot be used outside of a transaction.`
This depends on the fact that in DatabaseBackend.on_chord_part_return a transaction.atomic is created on the default database and not on the one associated with ChordCounter.

We write this simple fix for handle also this case.

Setting we configured:

```
DATABASES = {
    "default": {
        "ENGINE": "django.db.backends.postgresql",
        "NAME": "data",
    },
    "celery_result": {
        "ENGINE": "django.db.backends.postgresql",
        "NAME": "results",
    },
}
DATABASE_ROUTERS = [
    "app.db_routers.DjangoCeleryResultRouter",
]
```

Router class:

```
class DjangoCeleryResultRouter:
    route_app_labels = {"django_celery_results"}

    def db_for_read(self, model, **hints):
        if model._meta.app_label in self.route_app_labels:
            return "celery_result"
        return None

    def db_for_write(self, model, **hints):
        if model._meta.app_label in self.route_app_labels:
            return "celery_result"
        return None

    def allow_relation(self, obj1, obj2, **hints):
        if obj1._meta.app_label in self.route_app_labels or obj2._meta.app_label in self.route_app_labels:
            return True
        return None

    def allow_migrate(self, db, app_label, model_name=None, **hints):
        if app_label in self.route_app_labels:
            return db == "celery_result"
        return None
```

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -246,7 +246,7 @@ class DatabaseBackend(BaseDictBackend):
         if not gid or not tid:
             return
         call_callback = False
-        with transaction.atomic():
+        with transaction.atomic(using=ChordCounter.objects.db):
             # We need to know if `count` hits 0.
             # wrap the update in a transaction
             # with a `select_for_update` lock to prevent race conditions.
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -7,7 +7,7 @@ from celery.exceptions import ChordError
 from celery.result import GroupResult, allow_join_result, result_from_tuple
 from celery.utils.log import get_logger
 from celery.utils.serialization import b64decode, b64encode
-from django.db import connection, transaction
+from django.db import connection, router, transaction
 from django.db.utils import InterfaceError
 from kombu.exceptions import DecodeError
 
@@ -246,7 +246,7 @@ class DatabaseBackend(BaseDictBackend):
         if not gid or not tid:
             return
         call_callback = False
-        with transaction.atomic(using=ChordCounter.objects.db):
+        with transaction.atomic(using=router.db_for_write(ChordCounter)):
             # We need to know if `count` hits 0.
             # wrap the update in a transaction
             # with a `select_for_update` lock to prevent race conditions.
--- a/t/unit/backends/test_database.py
+++ b/t/unit/backends/test_database.py
@@ -782,10 +782,11 @@ class test_DatabaseBackend:
         request.group = uuid()
         self.b.on_chord_part_return(request=request, state=None, result=None)
 
-    @pytest.mark.django_db(databases=['default', 'read'])
     def test_on_chord_part_return_multiple_databases(self):
-        """Test if the ChordCounter is properly decremented and the callback is
-        triggered after all chord parts have returned with multiple databases"""
+        """
+        Test if the ChordCounter is properly decremented and the callback is
+        triggered after all chord parts have returned with multiple databases
+        """
         gid = uuid()
         tid1 = uuid()
         tid2 = uuid()
@@ -793,7 +794,9 @@ class test_DatabaseBackend:
         group = GroupResult(id=gid, results=subtasks)
         self.b.apply_chord(group, self.add.s())
 
-        chord_counter = ChordCounter.objects.using("read").get(group_id=gid)
+        chord_counter = ChordCounter.objects.using(
+            "secondary"
+        ).get(group_id=gid)
         assert chord_counter.count == 2
 
         request = mock.MagicMock()
@@ -817,7 +820,7 @@ class test_DatabaseBackend:
         self.b.mark_as_done(tid2, result, request=request)
 
         with pytest.raises(ChordCounter.DoesNotExist):
-            ChordCounter.objects.using("read").get(group_id=gid)
+            ChordCounter.objects.using("secondary").get(group_id=gid)
 
         request.chord.delay.assert_called_once()
 
--- a/t/unit/backends/test_database.py
+++ b/t/unit/backends/test_database.py
@@ -782,6 +782,45 @@ class test_DatabaseBackend:
         request.group = uuid()
         self.b.on_chord_part_return(request=request, state=None, result=None)
 
+    @pytest.mark.django_db(databases=['default', 'read'])
+    def test_on_chord_part_return_multiple_databases(self):
+        """Test if the ChordCounter is properly decremented and the callback is
+        triggered after all chord parts have returned with multiple databases"""
+        gid = uuid()
+        tid1 = uuid()
+        tid2 = uuid()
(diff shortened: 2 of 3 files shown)


django_celery_results/backends/database.py
```diff
@@ -246,7 +246,7 @@ def on_chord_part_return(self, request, state, result, **kwargs):
         if not gid or not tid:
             return
         call_callback = False
-        with transaction.atomic():
+        with transaction.atomic(using=ChordCounter.objects.db):
```
AllexVeldman, line 249: `BaseManager.objects.db` returns the `django.db.router.db_for_read()` which in your situation works since it returns the same db alias as the `db_for_write()`.

https://github.com/django/django/blob/2be37b253341cfd1f1363c533e6f896230f047a7/django/db/models/manager.py#L143

Consider using `django.db.router.db_for_write(ChordCounter)` instead as we will be updating the `ChordCounter.count` value within the transaction.

AllexVeldman (changes requested):
Interesting, I did not consider multi-db setups when I wrote this.

I think this does warrant a testcase or two, specifically the case where the read and write db are not the same.
Maybe you could pull some inspiration from https://github.com/celery/django-celery-results/blob/main/t/unit/test_models.py

AllexVeldman:
related: #422

t/unit/backends/test_database.py
```diff
@@ -919,3 +920,57 @@ def test_backend_result_extended_is_false(self):
...
+        self.app.conf.result_backend = (
+            'django_celery_results.backends:DatabaseBackend')
+        self.app.conf.result_extended = True
+        self.b = DatabaseBackend(app=self.app)
+
+    def test_on_chord_part_return_multiple_databases(self):
+        """
+        Test if the ChordCounter is properly decremented and the callback is
+        triggered after all chord parts have returned with multiple databases
+        """
```
AllexVeldman, line 940: As there is no DATABASE_ROUTERS defined, this test does not show the correct db is selected.
Actually I think due to the way Djangos DATABASES.TEST.MIRROR option works this test will never fail for the reasons it is created.

To make this test find anything, you might need to add a 3rd database in `t.proj.settings`, which does not mirror "default". Then configure a db router for this test to route this app to the new db.
Then you can assert in this test the ChordCounter exists when using the new db and does not exist when using "default" (or "secondary", should not matter).

t/unit/backends/test_database.py
```diff
@@ -919,3 +920,57 @@ def test_backend_result_extended_is_false(self):
...
+
+        chord_counter.refresh_from_db()
+        assert chord_counter.count == 1
+
+        self.b.mark_as_done(tid2, result, request=request)
+
+        with pytest.raises(ChordCounter.DoesNotExist):
+            ChordCounter.objects.using("secondary").get(group_id=gid)
+
+        request.chord.delay.assert_called_once()
```
AllexVeldman, line 976: In `t.proj.settings.py`, add this to the DATABASES:

```python
        'read-only': {
            'ENGINE': 'django.db.backends.postgresql',
            'HOST': 'localhost',
            'NAME': 'other',
            'USER': 'postgres',
            'PASSWORD': 'devpass',
            'OPTIONS': {
                'connect_timeout': 1000,
                'options': '-c default_transaction_read_only=on',
            },
            'TEST': {
                'MIRROR': 'default',
            },
        }
```

This will create a read-only "mirror" connection to the default database.

The you can add the router to route read/writes accordingly:

```suggestion
class DjangoCeleryResultRouter:
    route_app_labels = {"django_celery_results"}

    def db_for_read(self, model, **hints):
        """Route read access to the read-only database"""
        if model._meta.app_label in self.route_app_labels:
            return "read-only"
        return None

    def db_for_write(self, model, **hints):
        """Route write access to the default database"""
        if model._meta.app_label in self.route_app_labels:
            return "default"
        return None


class ChordPartReturnTestCase(TransactionTestCase):
    databases = {"default", "read-only"}

    def setUp(self):
        super().setUp()
        self.app.conf.result_serializer = 'json'
        self.app.conf.result_backend = (
            'django_celery_results.backends:DatabaseBackend')
        self.app.conf.result_extended = True
        self.b = DatabaseBackend(app=self.app)

    def test_on_chord_part_return_multiple_databases(self):
        """
        Test if the ChordCounter is properly decremented and the callback is
        triggered after all chord parts have returned with multiple databases
        """
        with self.settings(DATABASE_ROUTERS=[DjangoCeleryResultRouter()]):
            gid = uuid()
            tid1 = uuid()
            tid2 = uuid()
            subtasks = [AsyncResult(tid1), AsyncResult(tid2)]
            group = GroupResult(id=gid, results=subtasks)

            assert ChordCounter.objects.count() == 0
            assert ChordCounter.objects.using("read-only").count() == 0
            assert ChordCounter.objects.using("default").count() == 0

            self.b.apply_chord(group, self.add.s())

            # Check if the ChordCounter was created in the correct database
            assert ChordCounter.objects.count() == 1
            assert ChordCounter.objects.using("read-only").count() == 1
            assert ChordCounter.objects.using("default").count() == 1

            chord_counter = ChordCounter.objects.get(group_id=gid)
            assert chord_counter.count == 2

            request = mock.MagicMock()
            request.id = subtasks[0].id
            request.group = gid
            request.task = "my_task"
            request.args = ["a", 1, "password"]
            request.kwargs = {"c": 3, "d": "e", "password": "password"}
            request.argsrepr = "argsrepr"
            request.kwargsrepr = "kwargsrepr"
            request.hostname = "celery@ip-0-0-0-0"
            request.properties = {"periodic_task_name": "my_periodic_task"}
            request.ignore_result = False
            result = {"foo": "baz"}

            self.b.mark_as_done(tid1, result, request=request)

            chord_counter.refresh_from_db()
            assert chord_counter.count == 1

            self.b.mark_as_done(tid2, result, request=request)

            with pytest.raises(ChordCounter.DoesNotExist):
                ChordCounter.objects.get(group_id=gid)

            request.chord.delay.assert_called_once()
```

These changes will have the test fail in the current state of this branch with:
```
django.db.transaction.TransactionManagementError: select_for_update cannot be used outside of a transaction
```
Which is what this PR is trying to solve.

Change to `using=router.db_for_write()` in database.py and the test will pass.

Note that this test will only fail on Postgres, but since select_for_update does not work for sqlite and CI runs with postgres this should not be  an issue.

dvdria2: Thank you @AllexVeldman

AllexVeldman (approved):

auvipy (approved):

auvipy:
thanks both of you

auvipy:
Yes

Issue #286, opened 2022-02-08
PENDING Tasks not added to database


urzbs:
Hi,

new tasks that have been pushed with .delay(*args)
do not show up as PENDING in the django-celery-results database (when all workers are offline)

(SUCCESS / FAILURE works just fine)

is this a misconfiguration on my side? I am kinda confused by this.

settings.py

```python
##### CELERY ######
CELERY_BROKER_URL = 'redis://localhost:6379'
CELERY_RESULT_BACKEND = 'django-db'
CELERY_ACCEPT_CONTENT = ['application/json']
CELERY_TASK_SERIALIZER = 'json'
CELERY_RESULT_SERIALIZER = 'json'
CELERY_TIMEZONE = 'Europe/Berlin'
CELERY_IMPORTS = ("myapp.tasks", )
CELERY_TASK_TRACK_STARTED = True

# celery setting.
CELERY_CACHE_BACKEND = 'default'

# django setting.
CACHES = {
    'default': {
        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
        'LOCATION': 'my_cache_table',
    }
}
```

celery.py

```python
import os
from celery import Celery

# Set the default Django settings module for the 'celery' program.
os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'myproject.settings')

app = Celery('myproject')

# Using a string here means the worker doesn't have to serialize
# the configuration object to child processes.
# - namespace='CELERY' means all celery-related configuration keys
#   should have a `CELERY_` prefix.
app.config_from_object('django.conf:settings', namespace='CELERY')

# Load task modules from all registered Django apps.
app.autodiscover_tasks()

```

tasks.py

```python
from celery import shared_task

@shared_task(queue='queue1')
def add(x, y):
    return x + y

```

push task
```python
from myapp.task import add
add.delay(1,2)
```

start worker:

```
celery -A myproject worker -l INFO -Q queue1
```

urzbs:
As a Workaround, i tried writing a wrapper so when tasks are being pushed to the broker; we manually add the TaskResult object with status=PENDING

this is "kinda" a workaround, but it is problematic to get the tasks name from the AsyncResult.

For this I have seen that there is a celery option mentioned:
``result_extended = True`` that should save more information it to the AsyncResult._meta

![result_extended](https://docs.celeryproject.org/en/master/userguide/configuration.html#result-extended)

However, the task name is still nowhere to be found in the AsyncResult object

[...]

PR #432, opened 2024-06-08
Store current time when TaskResult started

yktakaha4:
Thank you to all maintainers of this great OSS.

Referring to the issue below, I generate TaskResult when Task is `PENDING`.
ref: https://github.com/celery/django-celery-results/issues/286#issuecomment-1789094153

This code works fine, but I can't use `date_created` as the start time of the Task because the TaskResult already exists when `STARTED` is recorded.
Additionally, in my use case, I want to measure the time from when a task is added to the message broker until the worker starts processing it.

By recording `date_started` as the time associated with `STARTED`, we can calculate the waiting time in the queue or the actual processing time of the worker.
Use `django.db.models.functions.Now` to get the current time to avoid being affected by the server's time settings.
ref: https://docs.djangoproject.com/en/5.0/ref/models/database-functions/#now

This PR is backward compatible because it simply adds a column and sets a value.
For users who do not create TaskResults themselves, `date_created` and `date_started` have the same meaning.

### Testing

We can check the value of `date_started` in django admin.

admin

For testing, I implemented a 5 second sleep on `@shared_task`.
The time increases in the order `date_created < date_started < date_done`.

```sql
select id, status, date_created, date_started, date_done
from django_celery_results_taskresult
where id = 35;
```

| id | status | date\_created | date\_started | date\_done |
| :--- | :--- | :--- | :--- | :--- |
| 35 | SUCCESS | 2024-06-08 06:14:37.631030 +00:00 | 2024-06-08 06:14:37.650764 +00:00 | 2024-06-08 06:14:43.334453 +00:00 |

Please let me know if there is anything else that should be tested.


auvipy (commented):
I like the intent of the PR. however I need more through review before merge it

auvipy:
in the mean time, can you please review this PR? https://github.com/celery/django-celery-results/pull/407

yktakaha4:
@auvipy  Thank you for your quick review.

I have confirmed the fix for #407.
I noticed that no tests were written and no documentation was provided, but I couldn't think of any points that would functionally conflict with my PR.
(Should I point this out in the review?)

In my personal opinion, `TaskResult.state` is just for storing `celery.states`, and I thought it might make state customization possible in Celery and then change it, but I'm not confident.

docs/getting_started.rst
```diff
@@ -61,8 +61,8 @@ To use :pypi:`django-celery-results` with your project you need to follow these
 
         CELERY_RESULT_EXTENDED = True
 
-    If you want to track the execution duration of your tasks (by comparing `date_created` and `date_done` in TaskResult), enable the :setting:`track_started` setting.
-    
+    If you want to track the execution duration of your tasks (by comparing `date_started` and `date_done` in TaskResult), enable the :setting:`track_started` setting.
```
auvipy, line 64: can we think of any example, using the new feature here?

yktakaha4: @auvipy Added example. However, I'm not good at English, so I would like you to correct the sentence.

By the way, you previously asked whether this package should include the behavior of registering TaskResult at `states.PENDING` timing.
ref: https://github.com/celery/django-celery-results/issues/286#issuecomment-1279655622

If it is possible for me, I would like to create another PR. What do you think?

auvipy: you can certainly open another PR. but what are the issues with the open PR?

auvipy (approved):

docs/getting_started.rst
yktakaha4: Sorry I'm not good at explaining. There is nothing problems with this PR.
I would like to submit a new PR for the PENDING status registration.
If there is no problem with this PR, would you please merge it?

auvipy: you are most welcome to come with a new PR

Issue #376, opened 2023-03-24
Periodic task name is missing in TaskResult


Mapiarz:
Packages: celery 5.2.7, django-celery-beat 2.5.0, django-celery-results 2.5.0, django 3.2.18. Backed by postgres and rabbitmq.

Yes, I have set `CELERY_RESULT_EXTENDED` to `True` and other metadata is there, but not periodic task name. Not a duplicate of https://github.com/celery/django-celery-results/issues/342 - the periodic task name is missing in both list and detail view, in fact, it's not stored in the DB in the first place.

Any ideas?

cveilleux:
I am seeing the same behavior after updating to 2.4.0 from 2.2.0, task names are missing from admin and database.

Thanks for pointing me to the undocumented (?) breaking change regarding result_extended, for me it was the solution.

I am setting the celery config directly, not using the `CELERY_XXX` django settings shim, although it should be equivalent.

```python
app = Celery("myproj")
app.conf.update(
    [...]
    result_extended=True,
)
```

Double-check your CELERY_RESULT_EXTENDED is working and that the config does get sent all the way to celery.

auvipy:
> Thanks for pointing me to the undocumented (?) breaking change regarding result_extended, for me it was the solution.
> 
> I am setting the celery config directly, not using the `CELERY_XXX` django settings shim, although it should be equivalent.

Would you mind what is missing on maintainers side which will help to get rid of this undocumented breaking change? any suggestion or help would be highly appreciated

cveilleux:
Thanks for asking and putting time to maintain this library.

The change was technically a fix, but I would argue that since it changes a core behavior of the library, it is also a breaking change.

Maybe add a note with a bold BREAKING prefix to the 2.4.0 changelog.

I did not intend to hijack @Mapiarz issue. He does mention he did set CELERY_RESULT_EXTENDED to true.

If the setting is indeed set properly and you are sure it gets passed down to the celery conf object, this could be a duplicate of https://github.com/celery/django-celery-results/issues/289

andrei-v1:
> Any ideas?

Are you using RabibitMQ or Redis as broker?
I had same problem when I used RabibitMQ, when I switched to Redis everything works fine as expected

dalmasonto:
Using `redis` solves the missing `periodic_task_name`

Mapiarz:
@cveilleux It probably has to do with eager mode, yes.

[...]

PR #445, opened 2024-10-08
Fix for missing periodic task name

ntindicator:
Fix for issue [#376](https://github.com/celery/django-celery-results/issues/376). The message broker Rabbitmq isn't picking up the `periodic_task_name` when it's passed as a keyword arg to `task.apply_async()` in django-celery-beat. If it's added to the `headers` in `apply_async()` it becomes available in the `request` object in django-celery-results. For this change to work the following [commit](https://github.com/celery/django-celery-beat/commit/3ab299e8f7d46fc95a22ac1a0cf823b1e7d8b009) also needs to be merged, otherwise it won't get passed by django-celery-beat. 

If the task isn't triggered by django-celery-beat, but is done with code, such as using `delay()` or `apply_async()`, it will be missing, unless it's added to the headers in [`apply_async()`](https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.apply_async). [delay()](https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.delay) doesn't support the extra options. E.g. When triggering a child task with `delay()` Periodic Task Name was missing in the Task Results. 

These changes haven't been tested with Redis or any other message broker, only rabbitmq. `CELERY_RESULT_EXTENDED` was set to `True` in the settings.
Tested with: 
celery==5.4.0 
django-celery-beat==2.7.0 
django-celery-results==2.5.1
Django==4.2.16
rabbitmq==4.0.2

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -87,8 +87,8 @@ class DatabaseBackend(BaseDictBackend):
             if task_kwargs is not None:
                 _, _, task_kwargs = self.encode_content(task_kwargs)
 
-            properties = getattr(request, 'properties', {}) or {}
-            periodic_task_name = properties.get('periodic_task_name', None)
+            periodic_task_name = getattr(request, 'periodic_task_name', None)
+
             extended_props.update({
                 'periodic_task_name': periodic_task_name,
                 'task_args': task_args,


Nusnus (dismissed):
Is it possible to add unit tests please?

django_celery_results/backends/database.py
```diff
@@ -87,8 +87,8 @@ def _get_extended_properties(self, request, traceback):
             if task_kwargs is not None:
                 _, _, task_kwargs = self.encode_content(task_kwargs)
 
-            properties = getattr(request, 'properties', {}) or {}
-            periodic_task_name = properties.get('periodic_task_name', None)
+            periodic_task_name = getattr(request, 'periodic_task_name', None)
```
auvipy, line 90: thanks for the fix, curious, if removing properties create any regression or not? can you please elaborate the change bit more?

auvipy (changes requested):

django_celery_results/backends/database.py
ntindicator: This change reduces complexity by directly accessing the `periodic_task_name` attribute from request instead of an additional lookup inside properties. If `request.periodic_task_name` exists, it retrieves the value directly; otherwise, it defaults to `None`. The main reason for the change is that by passing the name in the headers of `apply_async()`, it is no longer available in properties, and is instead available in request. Removing properties shouldn't create any regression because it was only being used for the purpose of retrieving the `periodic_task_name`, and isn't being returned by the `_get_extended_properties` function or accessed in any other part of the code.

auvipy (approved):

PR #458, opened 2025-02-16
Update Django/Python support

### LINKED ISSUE #420: Django 5 support
Does the latest version work with Django 5? Anybody tried it? Is there a release planned with Django 5 support?

browniebroke:
- [x] Add support for Python 3.12
- [x] Add support for Django 5.1
- [x] Add support for Python 3.13
- [x] Drop support for Python 3.7 and 3.8
- [ ] Drop support for Django 3.2 - 4.1

EDIT: perhaps duplicated with https://github.com/celery/django-celery-results/pull/456

Fix #420


requirements/test.txt
```diff
@@ -1,4 +1,3 @@
-case>=1.3.1
```
browniebroke, line 1: This one might deserve its own separate PR... Thoughts?

t/conftest.py
```diff
@@ -31,6 +39,117 @@ def app(celery_app):
...
+        >>>     patching('sys.platform', 'darwin')  # set concrete value
+        >>>     patching.setenv('DJANGO_SETTINGS_MODULE', 'x.settings')
+        >>>     # val will be of type mock.MagicMock by default
+        >>>     val = patching.setitem('path.to.dict', 'KEY')
+    """
+    return _patching(monkeypatch, request)
+
+
 @pytest.fixture(autouse=True)
 def test_cases_shortcuts(request, app, patching):
```
browniebroke, line 154: As far as I can tell, the `patching` fixture was only thing that was coming from the `case` dependency. I took the current implementation from Celery (found here https://github.com/celery/celery/pull/7077)

Not sure if duplicating these ~100 lines of code is the best way forward but that does the job

auvipy: we might be using that from celery instead of case if those are already available there

auvipy: or we can handle this in a separate PR

auvipy (changes requested):

t/conftest.py
browniebroke: The version from Celery is defined under the tests directory (`celery/t`), which not included in [the distributions](https://github.com/celery/celery/blob/64d750bb113310503018c3c43a04d333fc1d2859/setup.py#L139), so it's not available in the version we have installed... [I tried](https://github.com/celery/django-celery-results/pull/461) to install from the source distribution, which seems to have the `t` directory included, but couldn't get to import the fixture.

browniebroke: So I have 2 approaches:

1. Extract the commit as is from this PR https://github.com/celery/django-celery-results/pull/462 - just works but fixture is duplicated with Celery
2. Use the fixture from Celery https://github.com/celery/django-celery-results/pull/461 - not working right now, so I assume it needs more work in Celery... Perhaps move the fixture from `t` to under `celery` to be included in the distribution, but I'm not sure about the side effects of doing so.

Approach 1 is more practical, and can unblock this PR immediately. Approach 2 is more purist but needs more work and guidance from one of the Celery maintainers.

auvipy (changes requested):
thanks for this PR. I will take parts from this PR for keeping older versions for a new release, to let the old users easier path to upgrade

PR #465, opened 2025-03-05
Make app AutoField configurable on first deployment

browniebroke:
While not fully addressing the problem reported in #307, it switches new installations to use a `BigAutoField` for primary key. Existing installations won't see any changes unless they replay the migrations and to recreate the result table.


auvipy (commented):
we should also think about what if existing projects want to migrate to this

browniebroke:
> we should also think about what if existing projects want to migrate to this

Thta's true, but that is more complicated and I don't think I want to tackle migrating existing installations in this pull request. 

Such migration might not even be feasible in a database agnostic way, as writing a safe migration might need to use database specific features. The project is tested against Postgres only, but it doesn't seem to use any PG only features? Is it supposed to work on other databases backend supported by Django?

docs/configuration.rst
```diff
@@ -0,0 +1,19 @@
+Configuration
```
browniebroke, line 1: There is another setting `DJANGO_CELERY_RESULTS_TASK_ID_MAX_LENGTH` which should probably be added here but I'd rather do this separately if you don't mind.

auvipy:
what is your take on this PR? although it is not 100% design wise

browniebroke:
> we should also think about what if existing projects want to migrate to this

There was [an excellent talk at DjangoCon Europe](https://pretalx.evolutio.pt/djangocon-europe-2025/talk/DLUM9M/) about how to do this kind of migration safely at scale, and the TL;DR is:

1. Add new shadow column
2. Backfill the values
3. Swap the columns in a transaction

The 2nd step is particularly tedious and manual. It can take a long time and has some operational implications (creates dead tuples in PostgreSQL, themselves causing spikes in VACUUM)... 

IMO django-celery-results should document a recipe but this should be left out of the package. Perhaps we could link to the talk when it's out?

Issue #468, opened 2025-03-06
Add Celery Task `shadow_name` Property


andreangelucci:
Celery tasks have a property called `shadow_name` that helps identify task execution in logs, as executions typically share the same task names.

After running a Celery task, we can monitor the execution with the `TaskResult` object. However, when there are multiple executions of the same task that differ only by their arguments, identification becomes challenging.

Would it be feasible to add a `shadow_name` property to the `TaskResult` model and enable filtering through Django admin? Alternatively, we could prioritize the shadow name as the `task_name` if it has been defined.

Shadow name example:

```python
def shadow_name(task, args, kwargs, options):
    return f"sleep_{args[0]}_seconds"

@shared_task(name='sleep-example', shadow_name=shadow_name)
def sleep_example(seconds):
    sleep(seconds)
```

After running `sleep_example.delay(15)`, we can see:

![Image](https://github.com/user-attachments/assets/08c4e218-bfc2-4539-a032-947608822a28)

![Image](https://github.com/user-attachments/assets/bb6e71ad-f4bb-4c9f-94d4-07272efe113b)

![Image](https://github.com/user-attachments/assets/c94fb703-10fc-44c8-9d32-90160c8d45f2)

![Image](https://github.com/user-attachments/assets/9e18734a-ecdf-4e04-93dc-f3f11ddba823)

![Image](https://github.com/user-attachments/assets/fab7607c-09ae-40d0-b796-0643d68d13b4)

Celery documentation:
https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.shadow_name

PR #469, opened 2025-03-06
Set Task Shadow Names as the Primary Task Name

andreangelucci:
Celery tasks have the `shadow_name` property used to define a custom name for tasks.
When a task has multiple executions, it becomes challenging to find their task results as they share the same task names, and the task IDs are usually UUIDs. Shadow names are useful to filter and identify results for monitoring.

This PR sets the shadow name as the primary choice for the `TaskResult.task_name` property. If it's not set, the default `task_name` will be used.

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -88,12 +88,14 @@ class DatabaseBackend(BaseDictBackend):
                 _, _, task_kwargs = self.encode_content(task_kwargs)
 
             periodic_task_name = getattr(request, 'periodic_task_name', None)
+            shadow_task_name = getattr(request, 'shadow_task_name', None)
+            task_name = shadow_task_name or getattr(request, 'task_name', None)
 
             extended_props.update({
                 'periodic_task_name': periodic_task_name,
                 'task_args': task_args,
                 'task_kwargs': task_kwargs,
-                'task_name': getattr(request, 'task', None),
+                'task_name': task_name,
                 'traceback': traceback,
                 'worker': getattr(request, 'hostname', None),
             })


auvipy (commented):
thanks for opening the PR. I will need some thorough review and thoughts on this before proceed

auvipy (changes requested):
as you can see, all the CI builds are failing due to this change. so we need proper tests cases to validate and avoid regression

andreangelucci:
Hey, @auvipy , thanks for the feedback!
I fixed the failing unit tests by adding the missing `shadow` property to the mocks.

auvipy:
> This PR sets the shadow name as the primary choice for the `TaskResult.task_name` property. If it's not set, the default `task_name` will be used.

I am thinking about changing the behavior, if it has any negetive effect on existing projects?

andreangelucci:
> I am thinking about changing the behavior, if it has any negetive effect on existing projects?

@auvipy I agree, this will change the current behavior. The idea came up when I noticed that shadow names are the best way to add custom identification to a task, yet the task result lacks this information. I've added more details here: https://github.com/celery/django-celery-results/issues/468

Alternatively, we could create a new field in the model. What do you think about this?

auvipy (changes requested):
so our concerns here are, if this break existing behavior? and can we keep the old behavior as primary and the shadow as secondary? need more thoughts to reach on a consensus. I'm open to more discussions

andreangelucci:
@auvipy By default, tasks don't have shadow names. These changes will affect the current behavior when users define custom names for their tasks. Flower, for example, displays shadow names as the main names when defined, and we are applying the same strategy.

For alternatives, we can either create a new field for the Shadow Name or implement a new configuration, such as USE_SHADOW_AS_MAIN_NAME.

If you agree to move forward, I can open a PR implementing one of the alternatives. If this initial approach is acceptable, I can add some unit tests. Otherwise, we can close this PR.

Let me know your thoughts.

Thanks!

auvipy:
we have to halt this for sometime until everything seems more obvious to me

PR #471, opened 2025-03-29
update python and django versions

auvipy:


auvipy (commented):
tests are passing with new changes

PR #481, opened 2025-06-09
Add support for stamps

### LINKED ISSUE #480: Stamps not supported in database backend
I've been trying to add stamps to a complex task canvas, but I've found that they don't get stored in the results table with `result_extended` enabled. I've attempted to trace why this isn't working and it seems that django-celery-results just doesn't retrieve stamp data.

In `celery.backends.base.Base`, there is a `_get_result_meta()` method, which builds extended metadata if extended results are enabled. Among other things [on L494 this retrieves stamps and assigns them to the meta object](https://github.com/celery/celery/blob/088c39c0f78b23a9cdf8d1c9e265ea64d02cfd86/celery/backends/base.py#L494).

In django-celery-results, collecting extended result data is instead performed in the `django_celery_results.backends.database.DatabaseBackend` by the `_get_extended_properties()` method, which doesn't do anything to stamps. The `_get_meta_from_request()` method retrieves metadata from the request, but there's nothing to actually put stamps into the metadata.

I've been able to fix this in my project by subclassing DatabaseBackend and overriding `_get_meta_from_request()`:

```py
from celery.backends.base import get_current_task
from django_celery_results.backends.database import DatabaseBackend as _DatabaseBackend

class DatabaseBackend(_DatabaseBackend):
    def _get_meta_from_request(self, request=None):
        meta = super()._get_meta_from_request(request)
        request = request or getattr(get_current_task(), "request", None)

        if getattr(request, 'stamps', None):
            meta["stamped_headers"] = request.stamped_headers
            meta.update(request.stamps)

        return meta
```

It seems the cache backend does support stamps as it doesn't override the `_store_result()` method of `BaseKeyValueAttributeStore`

freddyheppell:
Fixes #480

Adds support for [stamping](https://docs.celeryq.dev/en/latest/userguide/canvas.html#stamping) by copying stamps into the result meta. This is the same approach Celery's base backend takes.

I wrote a test but it's currently marked as skipped, as I couldn't figure out how to set the stamps on the request. You can't set them directly on the `Request` object and I wasn't sure if I could pass something into the `_create_request` helper to set them.

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -111,7 +111,8 @@ class DatabaseBackend(BaseDictBackend):
         meta = getattr(request, "meta", {})
 
         if getattr(request, "stamps", None):
-            meta["stamped_headers"] = request.stamped_headers
+            if hasattr(request, "stamped_headers"):
+                meta["stamped_headers"] = request.stamped_headers
             meta.update(request.stamps)
 
         return meta


t/unit/backends/test_database.py
```diff
@@ -550,6 +550,37 @@ def test_backend__task_result_meta_injection(self):
         tr = TaskResult.objects.get(task_id=tid2)
         assert json.loads(tr.meta) == {'key': 'value', 'children': []}
 
+    @pytest.mark.skip("Need to find how to inject stamped headers")
```
auvipy, line 553: mmmm.... not sure if we should merge this with this test only.... should we wait bit more?

auvipy (changes requested):

t/unit/backends/test_database.py
freddyheppell: Yeah I agree this needs testing - I just couldn't work out how to actually go about setting the stamp on the request

auvipy: were you able to figure that out?

Issue #442, opened 2024-09-15
use JSONField instead of TextField


hosseinmp76:
is there any reason not to use jsonfield for fields with json value stored in them?

auvipy:
at that time there was no native jsonfield support

hosseinmp76:
Thanks for your reply.
Is backward compatibility an issue? If not, I would like to make a PR and change them if you wish.

auvipy:
it is and issue, but lets not limit that for any innovative solution, I request you to come up with a draft PR to showcase your improvement.

PR #487, opened 2025-07-10
Replace texfields with jsonfield

mumarkhan999:
Replacing TextField with JSONField.

**Related Issue**

- https://github.com/celery/django-celery-results/issues/442

### DIFF
--- a/django_celery_results/models.py
+++ b/django_celery_results/models.py
@@ -149,7 +149,8 @@ class ChordCounter(models.Model):
         help_text=_("Celery ID for the Chord header group"),
     )
     sub_tasks = models.JSONField(
-        default=None,
+        null=True,
+        default=list,
         help_text=_(
             "JSON serialized list of task result tuples. "
             "use .group_result() to decode"


mumarkhan999:
Hi @auvipy,
Could you please review this PR when you get a chance?

auvipy (changes requested):
why so many schema migration files? can we reduce them? Also the change is not backward compatible as far as I can understand

mumarkhan999:
Hi @auvipy,

There are total 4 migration files.
Two of them are `schema` migration files and other two are `data migration` files.
 ...
The first migration is to add a new `JSONField` while keeping the old `TextField` as it is
The second and third one are to copy data from `TextFields` to `JSONFields`.
And the last one is to remove the old `TextFields`
...
I opted this approach just to keep the old data safe.
The removal of prevous textfields will not be done unless the data has been successfully copied into jsonfield.

danihodovic:
I'd love this to get merged in to simplify derserialization of task results when performing manual retries (creating a new task with args, kwargs).

PR #497, opened 2025-12-10
save date_created as timezone-aware

### LINKED ISSUE #496: date_started is not timezone-aware
# Steps to Reproduce

 1. Set database server to a timezone that's not UTC
 1. Launch a job
 1. date_started is populated with a datetime based on the server's timezone

# Expected Behavior

date_started has a timezone attached

# Observed Behavior

date_started does not have a timezone attached, so it's interpreted as UTC when read back

icovada:
Closes:  #496

### DIFF
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -8,7 +8,6 @@ from celery.result import GroupResult, allow_join_result, result_from_tuple
 from celery.utils.log import get_logger
 from celery.utils.serialization import b64decode, b64encode
 from django.db import connection, router, transaction
-from django.db.models.functions import Now
 from django.db.utils import InterfaceError
 from kombu.exceptions import DecodeError
 
--- a/django_celery_results/backends/database.py
+++ b/django_celery_results/backends/database.py
@@ -15,6 +15,7 @@ from kombu.exceptions import DecodeError
 from ..models import ChordCounter
 from ..models import GroupResult as GroupResultModel
 from ..models import TaskResult
+from ..utils import now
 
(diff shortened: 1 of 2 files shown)


django_celery_results/backends/database.py
```diff
@@ -146,7 +146,7 @@ def _store_result(
         )
 
         if status == states.STARTED:
-            task_props['date_started'] = Now()
+            task_props['date_started'] = now()
```
auvipy, line 149: can you please also add unit test for verifying this?

auvipy (changes requested):

auvipy (approved):

auvipy (commented):
do we need to check if the schema migrations is needed for this change?

t/unit/backends/test_database.py
```diff
@@ -968,6 +968,26 @@ def test_custom_state(self):
...
+        self.b.mark_as_started(tid)
+
+        # Verify date_started is set and timezone-aware
+        tr = TaskResult.objects.get(task_id=tid)
+        assert tr.date_started is not None
+        assert isinstance(tr.date_started, datetime.datetime)
+        # Check if datetime is timezone-aware (has tzinfo set)
+        assert tr.date_started.tzinfo is not None
+        assert tr.date_started.tzinfo.utcoffset(tr.date_started) is not None
+
```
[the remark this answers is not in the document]
auvipy: @icovada can you please cross check this?

django_celery_results/backends/database.py
auvipy, line 149: I think we should also add tests to verify this change

auvipy (changes requested):

PR #505, opened 2026-04-04
Chore: Move static metadata to pyrpoejct.toml

dpgraham4401:
## Changes

This PR is a part of an incremental migration from `setup.py` to `pyproject.toml`, currently this PR adds all static metadata fields to but keeps the project dependencies as dynamic fields in `pyproject.toml` and use the current set up with `setup.py` and the existing requirements text files. 

- Move static project metadata fields from `setup.py`/`__init__.py` to `pyproject.toml`
- migrate from unmaintained `bumpmyversion` to fork `bump-my-version`
- update the `Makefile`
    - `make test` runs pytest directly
    - `make build` is a `PHONY` target and uses the PYPA's build package to pull date from both `setup.py` and `pyproject.toml`
    -


.bumpversion.cfg
```diff
@@ -1,17 +0,0 @@
-[bumpversion]
```
auvipy, line 1: this is wrong. please revert this. also it would be best to abandon this pr. there are many open bug reports, you can work on those.

dpgraham4401: > also it would be best to abandon this pr.

Got it. I'll abandon this PR since, based on this, I don't expect this PR will get merged.

> this is wrong. please revert this.

I've reverted it. But [`bumpversion`](https://github.com/peritus/bumpversion) hasn't been maintained for so long that it's successor project [`bump2version`](https://github.com/c4urself/bump2version?tab=readme-ov-file#bump2version) is _also_ no longer maintained. This PR would have moved the project to the grandchild successor project [`bump-my-version`](https://github.com/callowayproject/bump-my-version) which can be configured in toml format and doesn't need this file.

